PR #13, opened 2022-01-26
chore(Style): Styling nits

rudouglas:


src/pages/instant-observability.js
```diff
@@ -338,6 +338,8 @@ const QuickstartsPage = ({ data, location }) => {
                 --svg-color: var(--color-neutrals-700);
                 box-shadow: none;
                 max-width: 630px;
+                font-size: 14px;
```
aswanson-nr, line 341: I don't think this is affecting the placeholder value in the search bar, it's still 16px when I run locally

aswanson-nr (commented):
Overall looks good! Although we appear to be missing the changes to the `GuidedInstallTile` paragraph

src/pages/instant-observability.js
rudouglas: ahh yes that was being overwritten good catch

aswanson-nr (approved):

PR #16, opened 2022-02-01
fix: branch protection update

rudouglas:


aswanson-nr:
This is no longer an issue, closing.

PR #20, opened 2022-02-04
chore: removed build your own tile

jpvajda:
This PR removes the build your own tile and replaces the create issue button with build your own.

### DIFF
--- a/src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
+++ b/src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
@@ -86,7 +86,7 @@ const GlobalFooter = ({ className }) => {
           <Button
             as={ExternalLink}
             variant={Button.VARIANT.OUTLINE}
-            to="https://developer.newrelic.com/contribute-to-quickstarts/"
+            href="https://developer.newrelic.com/contribute-to-quickstarts/"
             instrumentation={{ component: 'GlobalFooter' }}
           >
             <Icon
--- a/src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
+++ b/src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
@@ -84,7 +84,7 @@ const GlobalFooter = ({ className }) => {
         </Link>
         <div>
           <Button
-            as={Link}
+            as={ExternalLink}
             variant={Button.VARIANT.OUTLINE}
             to="https://developer.newrelic.com/contribute-to-quickstarts/"
             instrumentation={{ component: 'GlobalFooter' }}
--- a/src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
+++ b/src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
@@ -103,6 +103,12 @@ const GlobalFooter = ({
               `}
             />
             Build your own
+            <Icon
+              name="external-link"
+              css={css`
+                margin-left: 7px;
+              `}
+            />
           </Button>
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
```diff
@@ -91,27 +83,26 @@ const GlobalFooter = ({
...
         <div>
-          {repository && (
-            <CreateIssueButton
-              pageTitle={pageTitle}
-              variant={Button.VARIANT.OUTLINE}
-              size={Button.SIZE.SMALL}
-              labels={issueLabels}
-              instrumentation={{ component: 'GlobalFooter' }}
+          <Button
+            as={Link}
```
aswanson-nr, line 87: Looks like this button is trying to do a relative link for some reason. I think this might need to be an `ExternalLink`
```suggestion
            as={ExternalLink}
```

jpvajda:
@aswanson-nr I tried changing to ExternalLink and the URL is still pointing to the I/O domain an not to the external link provided.

src/@newrelic/gatsby-theme-newrelic/components/GlobalFooter.js
```diff
@@ -91,27 +83,26 @@ const GlobalFooter = ({
...
-            <CreateIssueButton
-              pageTitle={pageTitle}
-              variant={Button.VARIANT.OUTLINE}
-              size={Button.SIZE.SMALL}
-              labels={issueLabels}
-              instrumentation={{ component: 'GlobalFooter' }}
+          <Button
+            as={ExternalLink}
+            variant={Button.VARIANT.OUTLINE}
+            to="https://developer.newrelic.com/contribute-to-quickstarts/"
```
zstix, line 89: According to [the docs](https://github.com/newrelic/gatsby-theme-newrelic/tree/main/packages/gatsby-theme-newrelic#externallink), I think we want to use `href` not `to`. Let's see if that solves the problem 🤷 
```suggestion
            href="https://developer.newrelic.com/contribute-to-quickstarts/"
```

jpvajda:
@zstix thanks for that suggestion! I did make that change, and still no luck. 🤔

aswanson-nr (changes requested):
The reason this link isn't working as intended is due to [this line](https://github.com/newrelic/gatsby-theme-newrelic/blob/ef77172fcd611007367769a52fe6872dda09bf4d/packages/gatsby-theme-newrelic/src/components/Link.js#L63-L68) in our `Link` component. The reason this is happening on this site is that our [site metadata](https://github.com/newrelic/instant-observability-website/blob/main/gatsby-config.js#L18) config in `gatsby-config.js` is still set to the developer site

jpvajda:
@aswanson-nr pointed to the need for the sitemeta data to be updated for the new domain, so I made that change and the external link now works 💯

src/pages/index.jsx
```diff
@@ -539,17 +539,6 @@ const QuickstartsPage = ({ data, location }) => {
...
-              id={RESERVED_QUICKSTART_IDS.BUILD_YOUR_OWN_QUICKSTART}
-              css={css`
-                background-color: var(--tertiary-background-color);
-              `}
-              href={QUICKSTARTS_REPO}
-              view={view}
-              logoUrl={BUILD_YOUR_OWN}
-              title="Build your own quickstart"
-              summary="Can't find a quickstart with what you need? Check out our README and build your own."
-            />
```
josephgregoryii, line 552: Are we going to bring back the Quickstart tile at a later point?

josephgregoryii (approved):
👍🏻  with a discussion comment about `QuickstartTile`

src/pages/index.jsx
josephgregoryii: If not, we should consider removing the file 😄

jpvajda:
@josephgregoryii I think we can leave it for now, as we are handing off this site to another team soon and it's possible they may want to be bring back the individual tile for building your own.

src/pages/index.jsx
aswanson-nr: The `QuickstartTile` component is used for all quickstarts! So this is all that needs to be done

jpvajda:
one closing comment, we did end up removing all the code for build your own, but it can be resurrected in git history if it's needed again.

PR #30, opened 2022-02-11
chore: CXPUI-7 removed feature tag from pack tile

RavitejaSurampudi:
removed featured tag from pack tile


aswanson-nr:
Closing and reopening to trigger builds now that they are available.

PR #38, opened 2022-02-18
Remove cookie based signup logic

aswanson-nr:
## Summary
* Remove logic to check for a cookie to determine if the install link
  should go to the platform or the signup url
* All install links will now go directly to the platform


nr-kkenney (approved):

jpvajda:
Our I/O Growth Leadership team, instructed us to keep this as is for now, and not change this to flow to point to login. so closing and @polfliet took an action to talk more about the ideal solution with AAA for next steps.

PR #50, opened 2022-02-28
feat: CXPUI-11 featured and most popular slide changes

RavitejaSurampudi:
1. removed featured categories from left navigation
2. added react slick for new carousels of most popular and feature
3. changed quick-start config for some tiles and raised pr (https://github.com/newrelic/newrelic-quickstarts/pull/870)
4. Now guided install tile will come as first tile in most popular carousel until search is empty state.
5. if the search box is not empty , then guided install tile will be the first one


src/components/IOBanner.js
```diff
@@ -1,12 +1,21 @@
-import React from 'react';
+import React, {useState} from 'react';
```
clarkmcadoo, line 1: It doesn't look like we are using `useState` so we don't have to import it here I don't think.

src/pages/index.jsx
```diff
@@ -180,15 +193,60 @@ const QuickstartsPage = ({ data, location }) => {
...
+    dots: false,
+    infinite: false,
+    speed: 300,
+    slidesToShow: 4,
+    slidesToScroll: 4,
+    adaptiveHeight:false,
+    adaptiveWidth:true,
+    responsive: [
+      {
+        breakpoint: 1024,
```
clarkmcadoo, line 209: We might consider making this `1081`, the breaking gets a little awkward between 820-1080 px due to the dynamic styles of the cards themselves.

Screen Shot 2022-02-28 at 10 43 30 AM

clarkmcadoo: Also, it would probably just be safe to make there only 1-2 slides at < 1081px widths

clarkmcadoo:
also, little nit-pick here, but it looks like we have a few imports in the `pages/index.jsx` lines 12, 15, 16 :)

src/components/IOBanner.js
RavitejaSurampudi: removed unused imports

src/pages/index.jsx
RavitejaSurampudi: updated break point settings

aswanson-nr (approved):
Code looks great and good to go!
I did notice some style issues on tablet sized screens (1600-1181px ish), screenshots are attached. Feel free to merge when ready!
Screen Shot 2022-03-01 at 12 53 13 PM
Screen Shot 2022-03-01 at 12 53 48 PM

zstix (approved):

Issue #46, opened 2021-12-06
[SSR Public Catalog] Create a 500 error page


aswanson-nr:
## Description
With the transition to using SSR for the I/O catalog pages, we now have the possibility to hit a 500 HTTP error on these pages.

### Design Questions

- [x] What is the preferred design for a [500 error](https://www.ionos.com/digitalguide/hosting/technical-matters/http-error-500-internal-server-error/) given we can load most of our components (see below) but not the navigation or search given that will be powered by the catalog service & SSR.
- [x] This is only for the I/O experience and doesn't really apply to the dev , docs or oss sites as those sites don't use the catalog service or SSR So we don't have to apply this solution to those sites.
- [x] Review language with UI-Writing team for suggestions.

### Design References 🎨 

*What components will load and fail to load in case of a 500*

<img width="1719" alt="Screen Shot 2021-12-07 at 4 13 14 PM" src="https://user-images.githubusercontent.com/1395158/145124375-6b121de9-b1c5-42ce-94ff-3c805035de31.png">

<img width="1437" alt="Screen Shot 2021-12-07 at 4 13 25 PM" src="https://user-images.githubusercontent.com/1395158/145121124-c2c7d9d2-4f40-43bc-aba7-6d1db3d8afd2.png">

### Mock Up 🧑‍🎨 

<img width="1677" alt="(I:O 500 error " src="https://user-images.githubusercontent.com/1395158/145863395-b6984f62-2a33-484d-a45c-ca4440818c34.png">

### Links
- [Gatsby 500 page docs](https://www.gatsbyjs.com/docs/how-to/adding-common-features/add-500-page/)

### Error message content

> We are currently experiencing technical issues loading the Instant Observability catalog. Please use the guided install tile below to start the install process. 

### Acceptance Criteria 

- [x] a design is provided that we can use
- [x] Ensure this approach works on mobile 
- [x] a 500 page is visible when running `yarn develop` at `localhost:8000/500`
- [ ] a 500 page is visible when the catalog service fails to load the navigation of I/O.

aswanson-nr:
@jpvajda and @legnaleama, are these mocks still good post redesign?

jpvajda:
👋  the last we spoke (if I recall)  it's totally fine to go with this for our initial approach as this page isn't expected to be visible often. we can always revisit the design in a second iteration as needed.

tabathadelane:
[...]

PR #58, opened 2022-03-02
feat: create 500 error page

tabathadelane:
Following the mockup in the ticket, this PR creates a 500 error page for the IO site. 

- It displays at `localhost:8000/500` with a `yarn develop`. I seems to work fine in mobile view.
- I am unsure how to test if it works when the catalog service fails. 

Screen Shot 2022-03-02 at 11 32 58 AM
Screen Shot 2022-03-02 at 11 52 37 AM

closes https://github.com/newrelic/instant-observability-website/issues/46


aswanson-nr:
Looks great! Although I'm not sure we want to keep the logo and text on the top left side, since that was removed during the redesign. @jpvajda What do you think?

jpvajda:
@aswanson-nr @tabathadelane asked that question and I suggested we keep it for context look up a few comments for my reason.

Issue #35, opened 2022-02-16
Layout shifts when clicking on Data sources tab on certain quickstarts


roadlittledawn:
## Summary

Layout seems to shift when clicking on Data sources tab on a quickstart details page.

Repro steps:

1. Go to a quickstart details page like [this one](https://newrelic.com/instant-observability/netlify/5861d7f5-26c0-43ad-bda8-c893c4b27b25) or this [one](https://newrelic.com/instant-observability/aws-s3/2e3b5255-667e-4084-8aad-3dd6e0df2f78)
2. Click on `Data sources` tab 
3. Watch content shift to the right.

### Screenshot

https://user-images.githubusercontent.com/2952843/154173925-9fb0f93d-808b-44cf-bfe7-c489bf7d477c.mp4

## Acceptance criteria
- [x] Clicking on Data sources tab maintains same layout like other tab views

roadlittledawn:
Seems to only happen on that netlify quickstart

jpvajda:
I also clicked on several quickstarts, it seems to be just this one. 🤔

jpvajda:
I also saw this on AWS S3

https://newrelic.com/instant-observability/aws-s3/2e3b5255-667e-4084-8aad-3dd6e0df2f78

PR #59, opened 2022-03-02
fix: quickstart detail layout shifts on some tabs

tabathadelane:
it was a semi-colon. 😃

Closes https://github.com/newrelic/instant-observability-website/issues/35


aswanson-nr (approved):
This makes total sense!

PR #72, opened 2022-03-09
Chore:hidding the carousel when we select any category

sjyothi54:
JIRA:https://newrelic.atlassian.net/browse/CXPUI-26
Description:When user selects a category in left navigation pane carousel of most popular and featured are supposed to be hidden
**OLD VIEW:**
when All option selected in left panel
image
when any category is selected
image
**NEW VIEW:**
when All option selected in left panel
image
when any category is selected
image

### DIFF
--- a/src/pages/index.jsx
+++ b/src/pages/index.jsx
@@ -432,94 +432,10 @@ const QuickstartsPage = ({ data, location }) => {
             </Overlay>
           </div>
 
-          {isSelectCategory && (
+          {isSelectCategory && isSearchInputEmpty && (
             <>
-              {isSearchInputEmpty && (
+              {mostPopularQuickStarts.length > 0 && (
                 <>
-                  {mostPopularQuickStarts.length > 0 && (
-                    <>
-                      <div
-                        css={css`
-                          --text-color: var(--primary-text-color);
-                          font-size: 16px;
-                          color: var(--color-neutrals-800);
-                          align-text: center;
-                          span {
-                            color: var(--text-color);
-                            /* target inner children of parent span */
-                            span,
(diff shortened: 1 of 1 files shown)


aswanson-nr:
Quick close/reopen to retrigger that build

src/pages/index.jsx
```diff
@@ -427,10 +432,94 @@ const QuickstartsPage = ({ data, location }) => {
             </Overlay>
           </div>
 
-          {isSearchInputEmpty && (
+          {isSelectCategory && (
```
josephgregoryii, line 435: I think we can consolidate these lines to just one logical expression

```suggestion
          {isSelectCategory && isSearchInputEmpty && (...)}
```

sjyothi54: @josephgregoryii Updated the changes

src/pages/index.jsx
```diff
@@ -427,7 +432,7 @@ const QuickstartsPage = ({ data, location }) => {
             </Overlay>
           </div>
 
-          {isSearchInputEmpty && (
+          {isSelectCategory && isSearchInputEmpty && (
```
aswanson-nr, line 435: This approach totally works, it just adds some complexity with a new hook. Another option would be to key off of the `category` hook, which is already being/unset by the existing code. Then you can completely remove the need for a new `isSelectCategory` hook.
```suggestion
          {category === '' && isSearchInputEmpty && (
```

aswanson-nr (approved):
Hey @sjyothi54! I noticed a more simple way of achieving this and left a comment, but that change is not required. Feel free to merge!

PR #75, opened 2022-03-10
chore:when we click on logo it has to open on same page

sjyothi54:
JIRA:https://newrelic.atlassian.net/browse/CXPUI-18
Description:In I/O Home page when we click on New Relic Logo it have to open  in same tab
@aswanson-nr / @jpvajda 
We changed externalLink component to anchor tag  because
if we add Changes to Link.js which is in this repo:https://github.com/newrelic/gatsby-theme-newrelic then changes will apply to the following sites's also
https://docs.newrelic.com/
https://developer.newrelic.com/
https://opensource.newrelic.com/


josephgregoryii (dismissed):

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
```diff
@@ -169,7 +169,7 @@ const GlobalHeader = ({ className, activeSite }) => {
               }
             `}
           >
-            <ExternalLink
+            <a
```
aswanson-nr, line 172: The only problem I can think of with using an `a` tag here instead of the `ExternalLink` component is that we'll lose out on some analytics that are built into the component. A quick fix for that would be to do what the `ExternalLink` component is doing and add a `useInstrumentedHandler` for the onClick event. [This](https://github.com/newrelic/gatsby-theme-newrelic/blob/7122abeea868758c89a0ca59b836f6af43b7d3e1/packages/gatsby-theme-newrelic/src/components/Link.js#L47-L53) is how the ExternalLink handles it, you'll just need to change the `origin` field

aswanson-nr (changes requested):
Small comment about adding instrumentation!

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
sjyothi54: @aswanson-nr I updated the changes
In order to work this merge this code
linked PR:https://github.com/newrelic/gatsby-theme-newrelic/pull/645 has to be merge first

zstix (dismissed):

aswanson-nr (commented):
Hey @sjyothi54! I think some things may have just gotten lost in context on this PR. I don't think we want to make that update to the `ExternalLink` and `Link` components. Since it would only be used in this particular case. What you're trying to do can totally be done with the current tools available, the main thing is just keeping to tessen instrumentation for that link. I put together a quick example in #98, which should cover all the use cases for this link. If that meets all your requirements, feel free to pull that code over here or just merge that PR!

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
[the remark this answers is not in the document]
sjyothi54: @aswanson-nr Yes i made the changes

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
```diff
@@ -169,8 +170,13 @@ const GlobalHeader = ({ className, activeSite }) => {
...
             `}
           >
-            <ExternalLink
-              href="https://newrelic.com/"
+            <a
+              href="https://newrelic.com"
+              onClick={useInstrumentedHandler(null, {
+                eventName: 'externalLinkClick',
+                category: 'LinkClick',
+                origin: 'globalHeader',
```
aswanson-nr, line 178:
```suggestion
                origin: 'globalHeader',
                href: 'https://newrelic.com',
```

aswanson-nr (commented):
Just needs a `href` attribute and it's good to go, I'll commit my suggested change and merge!

josephgregoryii (approved):

aswanson-nr (approved):

PR #101, opened 2022-03-14
Fetch quickstarts from catalog service

### LINKED ISSUE #95: [I/O Site] Update fetch quickstart workflow to pull from catalog service
### ISSUE

We disabled [Fetch Quickstarts](https://github.com/newrelic/instant-observability-website/actions/workflows/fetch-quickstarts.yml) to allow us to launch the [Carousel feature](https://github.com/newrelic/instant-observability-website/pull/50)  but there is hero requst to update the external site with cloud flare content (See [PR](https://github.com/newrelic/newrelic-quickstarts/pull/915/files))  as there is launch today with a partner. 

`as a note:` this workflow has been off for 13 days, and the SRR work would make this workflow irrelevant but we still have bugs to resolve before we can release SSR.

### Steps to reproduce 

look for cloudflare on the external site, then look for cloudflare in the internal catalog, notice it's only visible internally.

### Expected

We can see cloud flare on the external site. 

**can we either** 
1.  re-enable the workflow, so we can update the quickstart content. then disable it again.
2. Manually update the site to include cloudflare quickstart.

aswanson-nr:
# Summary
Have the `fetch-quickstarts` workflow pull from the nr1Catalog API instead of the current one. 

* Includes formatting code to transform quickstarts into a shape that this site supports
* defaults some values, such as `url` for components, to `null` since that field does not exist in the catalog API.
  * Note: we never used these fields anyway
* Update support level enum with new value for `NEW RELIC` level

Closes #95


LizBaker (approved):
🎉

nr-kkenney (approved):
cloudflare there. clicked a few links and quickstarts, did some searches, looks good.

josephgregoryii (approved):
Code looks solid and seeing the new cloudfare. Everything seems to be working in order

Issue #84, opened 2022-03-12
[I/O Site] "Install New Relic" button inside leaves Guided Install container


josephgregoryii:
## Summary

In the Guided Install tile, the "Install New Relic" button leaves the parent container.

## Steps to reproduce

1. Navigate to [I/O Site](https://newrelic.com/instant-observability/)
2. Filter the categories OR add input to the search query
3. "Install New Relic" button leaves Guided Install Tile

<details>
<summary>Reference of issue</summary>

<img width="648" alt="Screen Shot 2022-03-11 at 3 51 11 PM" src="https://user-images.githubusercontent.com/56952981/157993482-2885f0f9-3f3e-435f-a328-9b06e8a0c4dc.png">

</details>

Possible location of bug within the code:
https://github.com/newrelic/instant-observability-website/blob/6bb7998283909e42b9bd4608b3375ca4b06bb150/src/components/GuidedInstallTileMostPopular.js#L110-L123

## Acceptance Criteria

- [x] Locate cause of bug and fix the positioning to remain inside the Guided Install Tile
- [x] Filter categories through the sidebar and verify if the install new relic button leaves the container
- [x] Add an input in the search query and verify if the install new relic button leaves the container

PR #103, opened 2022-03-15
Install button stays in guided install tile

tabathadelane:
`Install New Relic` button was set to `position: absolute` in relation to the page `body`. Now it is set to its parent element. 

closes https://github.com/newrelic/instant-observability-website/issues/84


josephgregoryii (approved):
Seems to work as expected!

Issue #82, opened 2022-03-11
[I/O Site] No limit set on search input length


josephgregoryii:
## Summary

The search input for the I/O Site does not limit the characters to a specified amount. It allows the user to search for extremely long inputs, which leads to breaking the UI. Not limiting the search also allows for the input to overlap with the “x” button to close out search.

<details>
<summary>Example of long inputs</summary>

![Screen Shot 2022-03-11 at 11 32 38](https://user-images.githubusercontent.com/56952981/157988920-3b940dc1-d3b7-4927-a722-2225c0acdb31.png)

![Screen Shot 2022-03-11 at 11 47 42](https://user-images.githubusercontent.com/56952981/157988941-059214f8-580f-4bac-8311-35d8d95f849a.png)

</details>

## Acceptance Criteria
- [ ] Limit search input to 128 or less characters (Specific cap TBD)
- [ ] Test to make sure the search input gets capped to that query

jpvajda:
@tabathadelane thanks for taking this on! 🥳

PR #104, opened 2022-03-15
Long search strings no longer break UI

tabathadelane:
There was some debate about whether to limit the max number of characters, or clip extra long input with css. This handles hiding extra long strings in the `input` bar and in the `results` span. 

We can still set a limit on the characters if we want, but editing the `input` `maxLength` attribute I think would require shadowing the `SearchInput` component from the theme. This seemed simpler. 

Screen Shot 2022-03-15 at 4 17 56 PM

closes https://github.com/newrelic/instant-observability-website/issues/82


aswanson-nr (approved):
Looks good to me!

PR #118, opened 2022-03-25
fix-carousel-on-load/CXPUI-38

pkosanam:
OLD:

When page, loads user used to see 4 slides of the carousal
User is not able to see any label for the carousal
eg: headings named "Most Popular" and "Featured" are not displayed on mobile screens i.e. after "QUICKSTARTS_COLLAPSE_BREAKPOINT"(760px) resolution
NEW:

Fixed - we are now displaying a loader until the page completely loads
To show difference between two carousals, I have applied label on the mobile screens i.e. after "QUICKSTARTS_COLLAPSE_BREAKPOINT"(760px) resolution


pkosanam:
closing - as the branch name is not right.

Issue #99, opened 2022-03-14
[I/O Site] incorrect 404 page rendering on newrelic.com/instant-observability


jpvajda:
The [gatsby-theme 404](https://developer.newrelic.com/404) don't seem to actually be rendering  when a user hits a 404 page on [newrelic.com/instant-observability](https://newrelic.com/instant-observability/4040404).

nr-kkenney:
i started by investigating the broken 404 page since we want to make sure that page works correctly before we purposefully redirect to it.

if we visit a page that does not exist on our site, [for example](https://instantobservabilitywebsite.gatsbyjs.io/instant-observability/codestreams/29bd9a4a-1c19-4219-9694-0942f6411ce7), if we look at the console logs we will see:

```none
react-dom.production.min.js:216 TypeError: Cannot read properties of undefined (reading 'load')
    at tessen.js:76:21
    at H (tessen.js:38:10)
    at Module.z (on-route-update.js:5:3)
    at api-runner-browser.js:26:34
    at Array.map (<anonymous>)
    at n.h (api-runner-browser.js:17:25)
    at S (navigation.js:44:5)
    at n.a.componentDidMount (navigation.js:224:5)
    at gu (react-dom.production.min.js:219:411)
    at Mi (react-dom.production.min.js:259:160)
```

so it looks like failing to load tessen is the reason for the broken page. if we look at the network tab, we see:
```
Request URL: https://instantobservabilitywebsite.gatsbyjs.io/tessen.min-1.14.0.js
Request Method: GET
Status Code: 404
```
for the call which tries to get tessen. it turns out, by adding a path prefix for our site, that resource now lives at `https://instantobservabilitywebsite.gatsbyjs.io/instant-observability/tessen.min-1.14.0.js`.

as my next step, I tried to think about what we could do. it seems like we could:
1. move the resource to the right spot (root of the domain) while still using path prefix.
2. remove path prefix.
3. create a new / shadowed 404 page to look at the correct spot for tessen or not break when tessen isn't there.
4. update the theme so we can supply a path for where to look, or something similar (modify the 404 page, etc).

[...]

PR #135, opened 2022-04-05
fix for broken 404 page

nr-kkenney:
## Summary
Changes were made to theme to resolve errors around using tessen when it wasn't loaded. Additionally, using Gatsby's path prefix with tessen was also fixed. So the 404 page should be fixed in the case where tessen isn't loaded, but should also be fixed to work now when it wasn't previously.

Finally, with these tessen changes, we can remove the hardcoded script block we had to get around this problem since tessen is now correctly sourced and should no longer break pages if not loaded.

## Links
Relates to: #99

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@emotion/styled": "^11.3.0",
     "@mdx-js/mdx": "^1.6.19",
     "@mdx-js/react": "^1.6.19",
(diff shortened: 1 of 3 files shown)


josephgregoryii (changes requested):
checking the build - sorry for early approve

josephgregoryii (approved):

PR #136, opened 2022-04-08
chore: CXPUI 45 : Handle redirects for URL when quickstart name changes

dikshab0202:
JIRA: https://newrelic.atlassian.net/browse/CXPUI-45
Description: handle redirects for URLs when a quickstart name changes
Fix: added new redirect json file and createRedirects action to generate any redirects

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -4,7 +4,6 @@ const externalRedirects = require('./src/data/quickstart-redirects.json');
 
 exports.createPages = async ({ actions, graphql, reporter }) => {
   const { createPage, createRedirect } = actions;
-  console.log("createRedirect->",createRedirect);
   const result = await graphql(`
     query {
       allQuickstarts {
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -24,7 +24,6 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       fromPath: from,
       toPath: to,
       isPermanent: true,
-      redirectInBrowser: true,
     });
   });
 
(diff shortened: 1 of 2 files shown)


pkosanam (changes requested):
Please make sure the indentation on quickstart-redirects.json file is right

Hi @aswanson-nr, request you to kindly review this PR as we have changed the URLs in "to" and "from" 
Thank you.

aswanson-nr (approved):
Tested the redirect and it works perfectly!

gatsby-node.js
```diff
@@ -19,6 +19,15 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
     }
   `);
 
+  externalRedirects.forEach(({ from, to }) => {
+    createRedirect({
+      fromPath: from,
+      toPath: to,
+      isPermanent: true,
+      redirectInBrowser: true,
```
aswanson-nr, line 27: Let's remove this option, it makes Gatsby use meta redirects to change pages. Which can have a bad impact on SEO results. Just FYI, when working with non-browser redirects and Gatsby, the redirects will not work in development mode and PR previews. What happens is Gatsby Cloud picks up the generated `build/_redirects.json` file and uses that to do server side redirects, which only happens on production domains.
```suggestion
```

aswanson-nr (changes requested):
Noticed one small thing we need to change!

dikshab0202:
> Noticed one small thing we need to change!

Hi @aswanson-nr , updated the commit suggestion in PR. Thank you..

aswanson-nr (approved):

pkosanam (approved):
looks good

PR #137, opened 2022-04-21
feat: nr1620-brand-new-header

sjyothi54:
JIRA:https://issues.newrelic.com/browse/NR-1620
Description:
Header changes in Home and Landing pages
Implemented new header design with nav bar menu links and logo and bottom border line
Screenshorts:
Web view:
Screenshot 2022-04-21 at 12 38 48 PM
Mobile view:
Screenshot 2022-04-21 at 12 40 59 PM
Screenshot 2022-04-22 at 7 33 50 PM


src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
```diff
@@ -110,20 +106,13 @@ const GlobalHeader = ({ className, activeSite }) => {
         css={[
           css`
             position: relative;
-            z-index: 1;
+            top: 0;
+            z-index: 80;
+            border: 1px solid #83878b;
```
aswanson-nr, line 111: The header in the designs only has a border on the bottom. Updating this to use `border-bottom` should solve that.
```suggestion
            border-bottom: 1px solid #83878b;
```

aswanson-nr (changes requested):
Hey @sjyothi54! This looks great, I just noticed that the border is going around the whole header and not just the bottom. I added a suggestion to take care of that!

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
sjyothi54: @AlecIsaacson Actually I also thought the same but in the figma design they had give the entire border thats why i keep it as a border. I changed the border to border-bottom.

PR #140, opened 2022-04-26
Chore:NR-1644-Landing pages-web- comprehensive monitoring section

sjyothi54:
JIRA:https://issues.newrelic.com/browse/NR-1644
Description:Update fonts/ colours in each of these sections of
Why monitor 
What should you look for in a .NET monitor 
What’s included in this quick start
What makes this unique
which looks similar to the figma design
NOTE: font-family will reflect only when the PR:https://github.com/newrelic/instant-observability-website/pull/137 got merged
Screenshots:
Screenshot 2022-04-25 at 8 27 06 PM
Screenshot 2022-04-25 at 8 27 19 PM
Mobile view:
Screenshot 2022-04-25 at 8 30 25 PM

### DIFF
--- a/src/components/QuickstartOverview.js
+++ b/src/components/QuickstartOverview.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import { css } from '@emotion/react';
 import { quickstart } from '../types';
-import OverviewTile from './OverviewTile';
 import Markdown from './Markdown';
 
 const allowedElements = [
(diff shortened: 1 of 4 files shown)


src/components/QuickstartOverview.js
```diff
@@ -23,71 +22,50 @@ const allowedElements = [
...
-            h1,
             h2 {
-              font-size: 1.5em;
-              font-weight: 600;
+              font-size: 44px;
+              line-height: 50px;
+              color: #1d252c;
+              margin-top: 104px;
             }
             h3 {
```
aswanson-nr, line 34: Looking at the preview, the H3s don't seem to have the same font-weight as the figma design. I'm not sure if that's because this doesn't have the new fonts yet, but just wanted to call that out.

src/components/QuickstartOverview.js
```diff
@@ -23,71 +22,50 @@ const allowedElements = [
...
-            key={index}
-            title={doc.name}
-            description={doc.description}
-            tag="Doc"
-          />
-        ))}
-      </div>
       {quickstart.description && (
         <div
           css={css`
```
aswanson-nr, line 27: To avoid having to list the font color for each html tag, we could put the color up here and let it cascade to the other styles.
```suggestion
          css={css`
            color: #1d252c;
```

src/data/quickstarts.json
```diff
@@ -2900,59 +2900,6 @@
     ],
```
aswanson-nr, line 2900: This file is auto generated by the `Fetch Quickstarts` workflow, please remove this before merging otherwise we could run into some tough conflicts when we merge this feature branch back into `main`

aswanson-nr (changes requested):
Changes look good! I noticed some small styling differences compared to the figma design and called them out. I also noticed 2 files that we probably don't want to commit any changes to, `src/data/quickstarts.json` and `src/data/related-pages.json` both of these files are auto generated by workflows. So we could run into some interesting merge conflicts when we merge the feature branch back into `main`

sjyothi54:
@AlecIsaacson I am closing this PR because even i am reverting the changes of related-pages.json it is not reflecting here
So i raised new PR:https://github.com/newrelic/instant-observability-website/pull/143 regarding this issue.

PR #142, opened 2022-04-26
chore: NR 1647 - How to use this quickstart and Authors section

dikshab0202:
JIRA - https://issues.newrelic.com/browse/NR-1647
Description: How to use this quickstart section and LandingPageFooter Section
Created copy of RelatedResources component and modified it and also created two new components - one for 'LandingPageFooter' section and another for 'HowToUseQuickstart' .

**Note** Other Quickstart section is in different PRs. Once all will be merged, will able to see quickstart contents.

Screenshot 2022-04-26 at 1 19 16 PM

Screenshot 2022-04-26 at 1 19 47 PM

Screenshot 2022-04-26 at 1 18 48 PM
'


src/components/AuthorsCollaborate/Authors.js
```diff
@@ -0,0 +1,153 @@
+import React from 'react';
```
aswanson-nr, line 1: For future understanding, it could be worth renaming this component. Since `Authors` implies it just handles the author section and not the support, collaborate, and related resources sections.

src/components/SupportSection.js
```diff
@@ -11,6 +11,14 @@ import {
 const SupportSection = ({ supportLevel, onClick: onLinkClick }) => {
   const supportLink = (
     <Link
+      css={css`
+    color: #1D252C;
```
aswanson-nr, line 15: It could be worth breaking these colors out into CSS variables, that way we only set the hex value in one place.

aswanson-nr (commented):
Looks good! Just noticed a few small things that could be worth changing. 
I also noticed that the quickstart contents are not being displayed
Screen Shot 2022-04-26 at 11 56 22 AM
I know this feature branch is in active development, so if that's intentional, feel free to ignore me.

src/components/SupportSection.js
dikshab0202: Resolved : I added few variables in styles.scss and using those variables in the components for colors.

src/components/AuthorsCollaborate/Authors.js
dikshab0202: Resolved: renamed it as LandingPageFooter. Let me know if you have any other suggestion.

aswanson-nr (approved):

PR #146, opened 2022-04-28
feat: NR3156 - Existing layout is not working for big screen size

dikshab0202:
JIRA - https://issues.newrelic.com/browse/NR-3156
Description : Layout is not working for big screen sizes and some small fixes.
**Note**: Created duplicate of Layout component which is coming from gatsby to avoid margin and max-width.
Fix: Now new component will work for big screens.


aswanson-nr (commented):
@dikshab0202 Looks good to me! I just noticed that the layout now expands to the full width and I think the design kept the content within the same width as the header. If that's not the case, feel free to ignore me!

PR #152, opened 2022-05-03
Move font handling to one file

aswanson-nr:
To attempt to fix the build, it's failing on fonts currently, this moves all the font importing to one file.


aswanson-nr:
Hey @dikshab0202 and @sjyothi54, we ran into issues with the font when building the site on Gatsby Cloud. So we changed up how they are handled so that they work with the path prefix and removed the unicode characters from the filepath since they were being parsed differently on Gatsby Cloud vs local development.

josephgregoryii (approved):

PR #154, opened 2022-05-04
NR-3590: No dark mode

tabathadelane:
-remove all `dark-mode` css styles
-override `body class='dark-mode'`  that is set from theme if `darkMode=true` exists in local storage
     -override to `false` if so
-shadow `themes.js` from `gatsby-theme-newrelic` and delete `.dark-mode` styles
-hard code css for `.dark-mode` styles set within the `Tag` component to prevent 'flashing'  of dark mode colors on page render


aswanson-nr (approved):
Looks fantastic! I'm not seeing any dark mode flashes

PR #155, opened 2022-05-06
Landing page banner

aswanson-nr:
# Summary
* Sets up the landing page banner
* Unifies some of the brand colors into css variables
* Move layout handling to `QuickstartDetails.js` instead of its child components
* I did use a general component for the banner backgrounds, since the styling between the two pages diverges. If we make a third page in the future that uses the background, then it would be worth generalizing.


josephgregoryii (approved):
Looks A1 👍🏻

PR #168, opened 2022-05-10
Remove node sass in favor of `sass`

aswanson-nr:
# Summary
`node-sass` is deprecated and does not support the arm64 cpu architecture. This updates the site to use `sass` instead, which is the replacement for `node-sass` and it supports arm64 cpu architectures. 
TLDR: makes it work on M1 macs

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -116,7 +116,12 @@ module.exports = {
         },
       },
     },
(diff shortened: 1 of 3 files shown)


josephgregoryii (approved):
Happy to get this through 👍🏻

PR #169, opened 2022-05-10
NR-5627 Fix issue with banner height and margin

mickeyryan42:
https://issues.newrelic.com/browse/NR-5627

min-height to make the banner size dynamic seems to fix this issue.  Adding vw text sizes and media queries to adjust text based on screen size.

### DIFF
--- a/src/components/LandingBanner/BannerBackground.js
+++ b/src/components/LandingBanner/BannerBackground.js
@@ -9,20 +9,13 @@ const BannerBackground = ({ children }) => {
   return (
     <div
       css={css`
-        --banner-height: 430px;
-
         --left-margin: calc(50% - 50vw);
-
         width: 100vw;
         left: var(--left-margin);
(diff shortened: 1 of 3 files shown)


aswanson-nr (changes requested):
Hey @mickeyryan42! This looks great and seems to totally fix the issue on most quickstarts, however I was able to find 2 that still break the page a bit.
1. 
Screen Shot 2022-05-10 at 2 11 32 PM

2. This one is probably going to be fixed by [NR-5626](https://issues.newrelic.com/browse/NR-5626), so we shouldn't need to account for such a widescreen image normally. We will just want to test it after that ticket is complete.
Screen Shot 2022-05-10 at 2 11 45 PM

aswanson-nr:
This requires some work from #170, so let's wait and merge this after that goes in.

aswanson-nr (approved):
Lookin' good!!!

PR #163, opened 2022-05-10
Update sign-up buttons in footer

zstix:
## Description
This will update the buttons in the "Get started today" footer to use the semantic-token-style class names and colors for the "primary" and "primary-accent" buttons.

Previously, the buttons were a slightly different color. Furthermore, they didn't have the correct hover treatment.

## Related Issue(s)
* NR-5642

## Screenshot(s)
Screen Shot 2022-05-09 at 16 57 39

PR #170, opened 2022-05-10
NR-5626: Landing banner screenshot size inconsistencies

tabathadelane:
- The banner now checks if there is a dashboard screenshot that fits in the desired aspect ratio for the design (figma shows ~1.97)
- if it doesn't have one, use the default image
- the provided default image is NOT the correct aspect ratio, so padding will be added if that image is chosen for display.

- This only loops over the first dashboard's screenshots to avoid nested loops, but that can be easily added if we want to be that motivate to find a screenshot 
- The aspect ratio range can also easily be changed depending on how varied we want the images to be displayed in the banner design. This currently looks for images between a 2.2 and 1.6 aspect ratio.

### Cloudflare does not have any screenshots that fit the aspect ratio range
Screen Shot 2022-05-10 at 4 03 31 PM

### Node.js does have an image within the desired aspect ratio
Screen Shot 2022-05-10 at 4 03 54 PM


src/components/GetStartedFooter.js
```diff
@@ -93,6 +93,9 @@ const GetStartedFooter = () => {
           css={css`
             background-color: var(--nr-black);
             color: var(--nr-green);
+            &:hover {
+              color: var(--nr-green);
+            }
```
aswanson-nr, line 98: I'm pretty sure @zstix handled this in #163, merging in the feature branch should take take of that.

src/components/GetStartedFooter.js
```diff
@@ -104,6 +107,9 @@ const GetStartedFooter = () => {
             background-color: var(--nr-green);
             color: var(--nr-black);
             border: 1px solid var(--nr-black);
+            &:hover {
+              color: var(--nr-black);
+            }
```
aswanson-nr, line 112: Same with this one

src/components/LandingBanner/index.js
```diff
@@ -168,6 +172,57 @@ const LandingBanner = ({ quickstart, className, location }) => {
...
+    });
+    return { width, height };
+  };
+
+  const isImgAspect16by9 = async () => {
+    let image = {
+      src: bannerImg.src,
+      style: bannerImg.style,
+    };
+    for (const screenshot of quickstart.dashboards[0].screenshots) {
```
aswanson-nr, line 203: I'm curious what kind of behavior this will have for quickstarts that don't have any dashboards, we might need to check for that first.

src/components/QuickstartDashboards.js
```diff
@@ -77,24 +77,34 @@ const QuickstartDashboards = ({ quickstart }) => (
           <Slider {...settings}>
             {dashboard.screenshots.map((imgUrl) => {
               return (
-                <div>
+                <div
```
aswanson-nr, line 80: I think these changes might just be appearing because this was branched off an older version, getting the latest feature branch should remove them

src/components/WhatsIncluded/Dashboards.js
```diff
@@ -32,8 +32,8 @@ const Dashboards = ({ quickstart }) => {
           css={css`
             display: inline-block;
             background: var(--background-grey-color);
-            margin: 0px 8px;
-            padding: 1px 4px;
+            margin: 0px 0px;
```
aswanson-nr, line 35: I think these also have to do with an older version of the feature branch, should be handled after an update.

aswanson-nr (changes requested):
This works so well!!! Just noticed a few small things to clean up, but then it should be good to go!

src/components/LandingBanner/index.js
tabathadelane: good catch! updated.

src/components/LandingBanner/index.js
```diff
@@ -168,6 +172,58 @@ const LandingBanner = ({ quickstart, className, location }) => {
...
+    };
+
+    for (const screenshot of quickstart.dashboards[0]?.screenshots ?? []) {
+      const { width, height } = await getURLMeta(screenshot);
+      const aspectRatio = width / height;
+      if (aspectRatio > 1.6 && aspectRatio < 2.2) {
+        //set quickstartImgUrl to this screenshot if its the aspect ratio fits the page layout
+        //the ideal ratio is near(~1.9)
+        image.src = screenshot;
+        console.log('new image', image);
```
aswanson-nr, line 211:
```suggestion
```

aswanson-nr (changes requested):
This looks ready to go! Just need to remove a console.log

src/components/LandingBanner/index.js
```diff
@@ -168,6 +172,57 @@ const LandingBanner = ({ quickstart, className, location }) => {
   );
 };
 
+function useDetermineBannerImg(quickstart, defaultImage) {
+  const defaultBannerImg = {
+    src: defaultImage,
+    //style the default img's element to fit 1.9 aspect ratio
+    style: { padding: ' 20px 100px 0' },
```
aswanson-nr, line 179: Let's edit the image to make it the correct aspect ratio, instead of having to deal with extra styling here.

aswanson-nr (changes requested):
Looks like we just want to get the default image into the correct aspect ratio before merging!

aswanson-nr (approved):

PR #173, opened 2022-05-12
Fetch files with Github API

josephgregoryii:
### Summary
- Preview page hit's a pull request via query parameter and fetches SHA
- Once the SHA is retrieved, we use the SHA to walk through the file tree and grab all files in quickstart directory


src/pages/preview.jsx
```diff
@@ -0,0 +1,60 @@
+import React, { useEffect } from 'react';
+
+import PropTypes from 'prop-types';
+
+const GITHUB_API_BASE_URL =
+  'https://api.github.com/repos/newrelic/newrelic-quickstarts/contents';
+
+const GITHUB_API_PULL_URL =
+  'https://api.github.com/repos/newrelic/newrelic-quickstarts/pulls';
```
josephgregoryii, line 9: We'll eventually move these constants and other helper functions to a designated file

src/pages/preview.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Recursive function to walk the file system in GitHub
+ * @param {String} url - GitHub API URL to walk the file system
+ * @returns {Array} array of files
+ **/
+const iterateDirs = async (files, url) => {
```
aswanson-nr, line 16: I think we it would be worth moving this to a more functional paradigm. That is mainly to stay in line with how React does things and make it easier to reason about the code when you're reading it. 
By that, I mean at this point in the codebase I have no real idea what `files` is or that it should be an empty array on the first call. Mainly, you have to go find where this function is called to fully understand what it's doing. If you compare that to something like this:
```js
const iterateDirs = async (url) => {
  const branchResponse = await fetch(url);
  const branchResponseJSON = await branchResponse.json();
  let fileAggregator = [];

  for (const dirListing of branchResponseJSON) {
    if (dirListing.type === 'dir') {
      const dirFiles = await iterateDirs(dirListing.url);
      fileAggregator = [...fileAggregator, ...dirFiles];
    }
    else {
      fileAggregator = [...fileAggregator, dirListing];
    }
  }

  return fileAggregator;
}
```
All you need to know is that it takes in a Github api url and returns an array of files.
I went super functional with this, but it would be fine using `arr.push` as well.

src/pages/preview.jsx
```diff
@@ -0,0 +1,60 @@
...
+  for (const dirListing of branchResponseJSON) {
+    if (dirListing.type === 'file') {
+      files.push(dirListing);
+    } else {
+      await iterateDirs(files, dirListing.url);
+    }
+  }
+};
+
+const fetchData = async (prNumber, quickstartPath) => {
```
aswanson-nr, line 29: I think giving this a more descriptive name would help out, maybe something like
```suggestion
const getQuickstartFilesFromPR = async (prNumber, quickstartPath) => {
```

src/pages/preview.jsx
```diff
@@ -0,0 +1,60 @@
...
+    `${GITHUB_API_BASE_URL}/quickstarts/${quickstartPath}?ref=${branchSHA}`
+  );
+  console.log(files);
+};
+
+const PreviewPage = ({ location }) => {
+  useEffect(() => {
+    const urlParams = new URLSearchParams(location.search);
+    const prNumber = urlParams.get('pr');
+    const quickstartPath = urlParams.get('quickstart');
```
aswanson-nr, line 48: I think we're missing a little logic here to redirect the user to the homepage when these aren't supplied. Currently I'm seeing a console error when they aren't added to the url.

src/pages/preview.jsx
```diff
@@ -0,0 +1,60 @@
...
+  const prResponseJSON = await prResponse.json();
+  const branchSHA = prResponseJSON.head.sha;
+
+  // Recursively walk the Github API from the root of the quickstart
+  const files = [];
+  await iterateDirs(
+    files,
+    `${GITHUB_API_BASE_URL}/quickstarts/${quickstartPath}?ref=${branchSHA}`
+  );
+  console.log(files);
```
aswanson-nr, line 41: For this to be 100%, I think we'd like to have this returning the files themselves. That way we can call this function "done" and not need more changes to it in future tickets.

aswanson-nr: I also noticed this is logging out the Github API representation of the file metadata, for this ticket I think we want to get to the point of logging out the raw file content itself.

aswanson-nr (commented):
Hey @josephgregoryii! I had one big comment on the recursive function, please let me know if that does/doesn't make any sense. The rest are just around finishing up the AC and some small naming comments!

src/pages/preview.jsx
josephgregoryii: This makes a lot of sense, especially if we want to reuse this for the future

src/pages/preview.jsx
```diff
@@ -60,7 +111,12 @@ const PreviewPage = ({ location }) => {
       return;
     }
 
-    getQuickstartFilesFromPR(prNumber, quickstartPath);
+    try {
+      getQuickstartFilesFromPR(prNumber, quickstartPath);
+    } catch (error) {
+      navigate('/');
+      return;
+    }
```
josephgregoryii, line 119: I put this here EOD with the assumption i will need to change this. Since `getQuickstartFilesFromPR` is async and inside a useEffect, I need a different approach to retrieve a `null` value from it to check for `response.status !== 200`

src/pages/preview.jsx
josephgregoryii: fixed 👍🏻

src/pages/preview.jsx
```diff
@@ -0,0 +1,172 @@
...
+  // Response containing files at root of PR
+  const prResponseJSON = await prResponse.json();
+  const branchSHA = prResponseJSON.head.sha;
+
+  // Recursively walk the Github API from the root of the quickstart
+  const fileAggregator = await iterateDirs(
+    `${GITHUB_API_BASE_URL}/quickstarts/${quickstartPath}?ref=${branchSHA}`
+  );
+
+  const rawContent = await getRawContent(fileAggregator);
```
aswanson-nr, line 113: I know we want raw files for yaml and json, but do we want raw files for the images?

josephgregoryii: This makes a ton of sense. using `src` for `img` tags via url to raw content put it all together 👍🏻

src/pages/preview.jsx
```diff
@@ -0,0 +1,172 @@
...
+    // grab query parameters to determine if it is a local preview or
+    // PR preview
+    const urlParams = new URLSearchParams(location.search);
+    const prNumber = urlParams.get('pr');
+    const quickstartPath = urlParams.get('quickstart');
+
+    // check to make sure query parameters are set
+    // otherwise, return home
+    if (!prNumber || !quickstartPath) {
+      navigate('/');
```
aswanson-nr, line 134: Let's verify this works with the path prefix.

aswanson-nr (commented):
Saw one spot we should test out with path prefixes enabled!

src/pages/preview.jsx
josephgregoryii: @aswanson-nr I just checked using `yarn build:production` and `yarn serve:production`.

Navigated to `localhost:9000/instant-observability/preview` without any of the query parameters and it took me be back to `localhost:9000/instant-observability`

src/utils/preview/fetchHelpers.js
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Async function handles fetching changed files in pull request from Github
+ **/
+export const getQuickstartFilesFromPR = async (prNumber, quickstartPath) => {
+  // Hit the Github API for SHA that references the PR branch
+  const prResponse = await fetch(`${GITHUB_API_PULL_URL}/${prNumber}`);
+
+  if (prResponse.status !== 200 || !prResponse.ok) {
+    return null;
+  }
```
zstix, line 100: We should consider `throw`ing an error or, at very least, sending a message to the console to help us debug if something goes wrong.

src/utils/preview/fetchHelpers.js
```diff
@@ -0,0 +1,114 @@
...
+
+  // Response containing files at root of PR
+  const prResponseJSON = await prResponse.json();
```
zstix, line 103: Kinda a nitpick, but we may be adding unnecessary context to the names of these variables. Since there are no other responses or JSON in this code block, these variables _could_ be simplified:
* `prResponse` --> `response` (or `resp` / `res` to copy convention used elsewhere)
* `prResponseJSON` --> `json`

aswanson-nr (approved):

src/utils/preview/fetchHelpers.js
```diff
@@ -0,0 +1,114 @@
...
+ * Async function grabs the raw content from files
+ * @param {Array} fileAggregator - array of Github metadata objects
+ * @returns {Array<Object>} - array of objects containg raw content to parse
+ *
+ **/
+export const getRawContent = async (fileAggregator) => {
+  const rawContent = Promise.all(
+    fileAggregator.map(async (rawMetadata) => {
+      return determineContent(rawMetadata);
+    })
```
zstix, line 85: I think this could be simplified as:
```suggestion
    fileAggregator.map(determineContent)
```

src/utils/preview/fetchHelpers.js
```diff
@@ -0,0 +1,114 @@
...
+};
+
+/**
+ * Async function grabs the raw content from files
+ * @param {Array} fileAggregator - array of Github metadata objects
+ * @returns {Array<Object>} - array of objects containg raw content to parse
+ *
+ **/
+export const getRawContent = async (fileAggregator) => {
+  const rawContent = Promise.all(
```
zstix, line 82: Do we need to `await` the `Promise.all` here?

src/utils/preview/fetchHelpers.js
```diff
@@ -0,0 +1,114 @@
...
+
+  const filePath = path.split('quickstarts/').pop();
+
+  const rawContentObj = {
+    type,
+    filePath,
+    fileName,
+    content,
+  };
+  return rawContentObj;
```
zstix, line 72: Super nitpicky: I'm seeing that, in a few places, we define a variable and then return it right away. It might be simpler to simply return the object.

zstix (commented):
I had a couple of questions / comments. This is looking great tho!!

src/utils/preview/fetchHelpers.js
```diff
@@ -0,0 +1,114 @@
+import { GITHUB_API_BASE_URL, GITHUB_API_PULL_URL } from '../../data/constants';
+
+/**
+ * Recursive function to walk the file system in Github
+ * @param {String} url - Github API URL to walk the file system
+ * @returns {Array} array of files
+ **/
+export const iterateDirs = async (url) => {
+  const branchResponse = await fetch(url);
+  const branchResponseJSON = await branchResponse.json();
```
zstix, line 10: Oh one other comment I forgot to include. It looks like we're doing no error handling for these recursive API calls. This could be a pretty tricky thing to troubleshoot and we might want some logging here.

src/utils/preview/fetchHelpers.js
aswanson-nr: Since we are returning this entire array, when this is called the caller handles the await. Although, I'm not opposed to awaiting the content within this function.

src/utils/preview/fetchHelpers.js
aswanson-nr: Totally agree here, let's at least log out the error.

src/utils/preview/fetchHelpers.js
aswanson-nr: Good catch here, if even one of these calls fails than we can't reliably construct a quickstart. So we'll definitely want to be logging any errors.
This brings up another question, in the case of a failure like this, do we want to continue rendering and display placeholders for anything that we don't have or would it be better to display some sort of error message. I'm leaning toward logging any useful information, but continue rendering since the use case for the `local` option is for user's while they are developing quickstarts. So some files may be missing while they work on it.

src/pages/preview.jsx
```diff
@@ -0,0 +1,63 @@
+import React, { useEffect, useState } from 'react';
+
+import PropTypes from 'prop-types';
+import { getQuickstartFilesFromPR } from '../utils/preview/fetchHelpers';
+import { navigate } from 'gatsby';
+
+const PreviewPage = ({ location }) => {
+  const [contentFiles, setContentFiles] = useState([]);
+
+  // TODO: Make this into a custom hook to reduce useEffect usage
```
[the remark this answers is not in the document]
josephgregoryii: Might as well! 😄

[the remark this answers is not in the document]
aswanson-nr: @tabathadelane and I are currently pulling it out into a custom hook as part of 4041. So I'm fine with this going into the feature branch "as is"

src/utils/preview/fetchHelpers.js
josephgregoryii: @aswanson-nr and I discussed and we've added `throw`ing to handle this

aswanson-nr (approved):
Error handling looks good!

src/utils/preview/fetchHelpers.js
josephgregoryii: fixed 👍🏻

src/utils/preview/fetchHelpers.js
josephgregoryii: fixed 👍🏻

PR #184, opened 2022-05-16
NR-2231 Fetch files from localhost

mickeyryan42:
Fetching files from localhost
https://issues.newrelic.com/browse/NR-2231


src/hooks/useLocalhostQuickstart.js
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react';
+
+import { getQuickstartFilesFromLocal } from '../utils/preview/fetchHelpers';
+import { parseQuickstartFiles } from '../utils/preview/parseHelpers';
+import { navigate } from 'gatsby';
+
+const usePullRequestQuickstart = (location) => {
+  const [contentFiles, setContentFiles] = useState([]);
```
josephgregoryii, line 8:
```suggestion
  const [quickstart, setQuickstart] = useState([]);
```

I made the same comment in #192 - we might want to consider changing this to hold the state of the quickstart, since we dont need to worry about the files once they're parsed into a quickstart

src/hooks/useLocalhostQuickstart.js
```diff
@@ -0,0 +1,39 @@
...
+  useEffect(() => {
+    const urlParams = new URLSearchParams(location.search);
+    const port = urlParams.get('port') || 3000;
+
+    /*
+     * Async function to get quickstart files from local
+     * and set to state variable
+     **/
+    const fetchFiles = async () => {
+      const rawFileContent = await getQuickstartFilesFromLocal(port);
```
josephgregoryii, line 19: We need some error handling here since `getQuickstartFilesFromLocal` calls a function that can throw an error. I think a simple `try`/`catch` should be fine

src/hooks/useLocalhostQuickstart.js
```diff
@@ -0,0 +1,39 @@
...
+
+      // Error handling in the chance Github returns
+      // a non 200 status
+      if (rawFileContent === null) {
+        navigate('/');
+        return;
+      }
```
josephgregoryii, line 26: We can use this in the `catch` mentioned above.

josephgregoryii:
Looks great! just a few comments above :)

src/hooks/useLocalhostQuickstart.js
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react';
+
+import { getQuickstartFilesFromLocal } from '../utils/preview/fetchHelpers';
+import { parseQuickstartFiles } from '../utils/preview/parseHelpers';
+import { navigate } from 'gatsby';
+
+const usePullRequestQuickstart = (location) => {
```
aswanson-nr, line 7: I think this is just a copy paste error, it's got the wrong hook name.
```suggestion
const useLocalhostQuickstart = (location) => {
```

src/hooks/useLocalhostQuickstart.js
```diff
@@ -0,0 +1,39 @@
...
+      setQuickstart(quickstart);
+    };
+
+    fetchFiles();
+  }, []);
+  
+  return quickstart;
+};
+
+export default usePullRequestQuickstart;
```
aswanson-nr, line 39: Same here
```suggestion
export default useLocalhostQuickstart;
```

src/hooks/usePullRequestQuickstart.js
```diff
@@ -0,0 +1,51 @@
...
+      navigate('/');
+      return;
+    }
+
+    /*
+     * Async function to walk the file system in Github
+     * and set the content to a stateful variable.
+     **/
+    const fetchFiles = async () => {
+      const rawFileContent = await getQuickstartFilesFromPR(
```
aswanson-nr, line 29: This might just be an artifact from pulling in the feature branch. But this needs a try catch around it.

aswanson-nr (commented):
Just found a few function names that need to be cleaned up!

src/hooks/useLocalhostQuickstart.js
mickeyryan42: Ahh good catch.  I just opened a new PR which doesnt have Tabitha's commit to keep it separate,  I'm going to update it on this one:
https://github.com/newrelic/instant-observability-website/pull/193

PR #186, opened 2022-05-17
fix/NR-9804: Hide carousel buttons if not needed

pkosanam:
**Description**:
1. Hide carousel arrows if there is no card to display.
2. Increase the gap above the lines 'Showing 464 results for:All quickstarts'. Gap should be in sync with the gap above the 'Featured' text above.

**Previous**:
image
image

**Now**:
image
image


mdumpati (approved):
approved

PR #192, opened 2022-05-17
NR-4041: Parse quickstart PR files

tabathadelane:
parses the raw files from the github API and shapes them to match what the `QuickstartDetails.js` expects

`http://localhost:8000/preview?pr=1090&quickstart=golden-signals-web` is a good url to use to look at both dashboard and alert content


src/hooks/usePullRequestQuickstart.js
```diff
@@ -0,0 +1,51 @@
+import { useEffect, useState } from 'react';
+
+import { getQuickstartFilesFromPR } from '../utils/preview/fetchHelpers';
+import { parseQuickstartFilesFromPR } from '../utils/preview/parseHelpers';
+import { navigate } from 'gatsby';
+
+const usePullRequestQuickstart = (location) => {
+  const [contentFiles, setContentFiles] = useState([]);
```
josephgregoryii, line 8:
```suggestion
  const [quickstart, setQuickstart] = useState({});
```
Since the hook has parsing now, we might want to change this to hold the state of the parsed quickstart instead since we don't need to worry about the files anymore.

src/hooks/usePullRequestQuickstart.js
```diff
@@ -0,0 +1,51 @@
...
+      // a non 200 status
+      if (rawFileContent === null) {
+        navigate('/');
+        return;
+      }
+
+      const quickstart = await parseQuickstartFilesFromPR(rawFileContent);
+
+      setContentFiles(quickstart);
+    };
```
josephgregoryii, line 44: We changed up how this function error handles. We add a try catch to handle that here:

https://github.com/newrelic/instant-observability-website/blob/962b6a8a3a8ccf47450819d95088c52801199f23/src/pages/preview.jsx#L38-L48

josephgregoryii: So we just need a try/catch here :)

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,99 @@
...
+const parseQuickstartDir = (file) => {
+  if (file.fileName.includes('logo') && file.type === 'image') {
+    parsedQuickstart.logoUrl = file.content;
+  }
+  if (file.type === 'yaml') {
+    //if config, parse the yaml
+    const loadYaml = yaml.load(file.content);
+
+    //itterate through the array of documentation objects to trim new lines
+    const trimmedDocumentation = loadYaml.documentation.map((doc) => {
```
josephgregoryii, line 20: `documentation` is an optional parameter and breaks on quickstarts that do not have a `documentation` field in their config file.

example: `http://localhost:8000/preview?pr=1060&quickstart=battlesnake`

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+};
+
+/**
+ * Async function handles parsing fetched files from pull request
+ * Transforms them into shape needed for QuickstartDetails.js
+ **/
+
+export const parseQuickstartFilesFromPR = async (rawFileContent) => {
```
josephgregoryii, line 94: I dont think we need async here, but i might be missing something!

josephgregoryii (commented):
This looks great! just a couple comments

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+  if (file.type === 'json') {
+    const dashboardContent = JSON.parse(file.content);
+    dashboardDirs[getDir]['name'] = dashboardContent.name ?? '';
+    dashboardDirs[getDir]['description'] = dashboardContent.description ?? '';
+  }
+  if (file.type === 'image') {
+    dashboardDirs[getDir]['screenshots'].push(file.content);
+  }
+};
```
josephgregoryii, line 61: Sorry! One last comment - I think I/O wants dashboards to be an array of objects.

example:
```
{
...,
dashboards: [
      {
        "name": "Battlesnake Performance",
        "description": "Monitor your Battlensakes performance and see how it stacks up against your competition!",
        "url": null,
        "screenshots": [
          "https://raw.githubusercontent.com/newrelic/newrelic-quickstarts/main/quickstarts/battlesnake/dashboards/battlesnake-performance/battlesnake-performance.png"
        ]
      },
    ...
  ]
}

```

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,99 @@
+import yaml from 'js-yaml';
+
+let dashboardDirs = {};
+let alertsDir = [];
+
+let parsedQuickstart = {
+  alerts: alertsDir ?? [],
+  dashboards: dashboardDirs ?? [],
+};
```
aswanson-nr, line 9: These variables are sitting in the global scope, let's get them moved into a function so that they are only scoped to that function and its return value. That will most likely require some changes to each function, I'll leave comments on each one.

src/utils/preview/parseHelpers.js
tabathadelane: i am trying to take care of that on line 99

`dashboardDirs = Object.values(dashboardDirs);`

but it's not running before its returning the quickstart 🥴

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,99 @@
...
+  const alert = {
+    details: loadYaml.description?.trim() ?? '',
+    name: loadYaml.name?.trim() ?? '',
+    type: loadYaml.type?.trim() ?? '',
+  };
+
+  alertsDir.push(alert);
+};
+
+const checkFileType = (rawFile) => {
```
aswanson-nr, line 79: This function name is a little misleading, it does check the file type, but it also parses the file as well. Since we're not exporting it, we can keep the name simple, maybe something like this:
```suggestion
const parseFiles = (rawFile) => {
```

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,99 @@
...
+    const file = rawFile[each];
+    if (file.filePath.includes('/dashboards/')) {
+      parseDashboardDir(file);
+    } else if (file.filePath.includes('/alerts/')) {
+      parseAlertsDir(file);
+    } else {
+      parseQuickstartDir(file);
+    }
+  }
+};
```
aswanson-nr, line 90: To remove the use of global variables, each of the parse functions could return their part of the quickstart object and then this function could handle stitching that together and returning it.

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,99 @@
...
+    name: loadYaml.name?.trim() ?? '',
+    type: loadYaml.type?.trim() ?? '',
+  };
+
+  alertsDir.push(alert);
+};
+
+const checkFileType = (rawFile) => {
+  for (const each in rawFile) {
+    const file = rawFile[each];
```
aswanson-nr, line 81: If you use a `for of` here, line 78 shouldn't be needed.
```suggestion
  for (const file of rawFile) {
```

aswanson-nr (commented):
I left a few comments around the globally scoped variables for the quickstart. I think we'll want to pull that scope out of global. Seems like a great opportunity to pair up and reword that part.

josephgregoryii (changes requested):
Code looks great! - Requesting changes as a placeholder to reference adding tests to this PR

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,157 @@
...
+  const parsedQuickstart = { ...quickstartDirs, ...quickstart };
+
+  return parsedQuickstart;
+};
+
+/**
+ * Async function handles parsing fetched files from pull request
+ * Transforms them into shape needed for QuickstartDetails.js
+ **/
+
```
zstix, line 145:
```suggestion
```

src/hooks/usePullRequestQuickstart.js
```diff
@@ -0,0 +1,52 @@
...
+     * and set the content to a stateful variable.
+     **/
+    const fetchFiles = async () => {
+      try {
+        const rawFileContent = await getQuickstartFilesFromPR(
+          prNumber,
+          quickstartPath
+        );
+
+        const parsedQuickstart = await parseQuickstartFilesFromPR(
```
zstix, line 35: This doesn't need to be async.

```suggestion
        const parsedQuickstart = parseQuickstartFilesFromPR(
```

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,157 @@
...
+      name: loadYaml.name?.trim() ?? '',
+      type: loadYaml.type?.trim() ?? '',
+    };
+    alerts.push(alert);
+  });
+
+  return alerts;
+};
+
+const parseFiles = (rawFile) => {
```
zstix, line 112: We probably want to name this argument `rawFiles` (note the plural) since this is supposed to be an array of files.

A docstring (or types) would really help here too. I'm not sure from looking at this if it's a file path, file content, or a buffer.

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,157 @@
...
+    dashboardFiles.length > 0 ? parseDashboardFiles(dashboardFiles) : [];
+  const alertsDir = alertFiles.length > 0 ? parseAlertFiles(alertFiles) : [];
+  quickstartDirs.dashboards = dashboardsDirs;
+  quickstartDirs.alerts = alertsDir;
+  const quickstart = parseQuickstartFiles(quickstartFiles);
+
+  //merge together the parsed quickstart content and the alerts/dashboard arrrays
+  const parsedQuickstart = { ...quickstartDirs, ...quickstart };
+
+  return parsedQuickstart;
```
zstix, line 138: Super minor, but we probably don't need to define a variable if we're just going to return it.

```suggestion
  return { ...quickstartDirs, ...quickstart };
```

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,157 @@
...
+    //parse and build alert object and add it to the array
+    const alert = {
+      details: loadYaml.description?.trim() ?? '',
+      name: loadYaml.name?.trim() ?? '',
+      type: loadYaml.type?.trim() ?? '',
+    };
+    alerts.push(alert);
+  });
+
+  return alerts;
```
zstix, line 109: Rather than making an empty array, looping over the array, pushing a modification into the empty array, and returning it, we could use [`Array.prototype.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) to simplify this:

```suggestion
  return alertFiles.forEach((file) => {
    const loadYaml = yaml.load(file.content);

    return {
      details: loadYaml.description?.trim() ?? '',
      name: loadYaml.name?.trim() ?? '',
      type: loadYaml.type?.trim() ?? '',
    };
  });
```

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,157 @@
...
+const parseDashboardFiles = (dashboardFiles) => {
+  const dashboards = {};
+  //split each filepath to get its dashboard dir
+  dashboardFiles.forEach((file) => {
+    const getDir = file.filePath.split('/dashboards/')[1].split('/')[0];
+    //add it to the dashboard dirs object as a key if new
+    //or update it if not
+    if (!dashboards[getDir]) {
+      dashboards[getDir] = { name: '', description: '', screenshots: [] };
+    }
```
zstix, line 81: I wonder if it would be simpler to get all the dashboard sub-directories first, and then loop through each to get the files. That way we don't need to work with an object or check if the key exists. Here's a rough sketch:

```js
// not sure if there's a built-in function from `fs` to get all subdirectories...
return getSubdirectories(dashboardFiles).map((directory) => {
  // find JSON file in directory
  // get name and description from JSON file
  // find all image files in directory
  return { name, description, screenshots };
});
```

That should produce the same results with a little less custom code. Totally a style choice tho. Your call.

src/utils/preview/parseHelpers.js
```diff
@@ -0,0 +1,157 @@
...
+
+const parseQuickstartFiles = (quickstartFiles) => {
+  let quickstartContent = {};
+  let config;
+
+  quickstartFiles.forEach((file) => {
+    if (file.type === 'yaml' && file.fileName.includes('config')) {
+      config = file;
+    }
+  });
```
zstix, line 31: This could probably be simplified to use [`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) here:

```js
  const config = quickstartFiles.find((file) => {
    return file.type === 'yaml' && file.fileName.includes('config');
  });
```

It could be further simplified to use [`path.basename`](https://nodejs.org/docs/latest-v17.x/api/path.html#pathbasenamepath-ext):

```js
  const config = quickstartFiles.find((file) => {
    return path.basename(file) === 'config.yml';
  });

File: src/utils/__tests__/parseHelpers.test.js [hunk 17]
```diff
@@ -0,0 +1,107 @@
...
+  dashboardContent,
+  expectedDashboardOutput,
+  installPlansOutput,
+  documentationInput,
+  documentationOutput,
+} = require('../mock_data/content');
+
+console.log(baseFiles);
+
+describe('Action: Parse helpers', () => {
```
aswanson-nr, line 19: I think the `Action:` part here was only used for tests that relate to Github actions, so let's just name it after the file it's testing.
```suggestion
describe('parseHelpers', () => {
```

aswanson-nr (approved):

PR #206, opened 2022-05-20
Add import for QUICKSTARTS_REPO

aswanson-nr:
# Summary
* QUICKSTARTS_REPO was not being imported here, this did not break
  because a quickstart will never not have a packUrl.
* Some formatting fixes as well


pkosanam (approved):
perfect.. 👍

PR #208, opened 2022-05-23
Quickstart Preview

josephgregoryii:


src/pages/preview.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+const PreviewPage = ({ location }) => {
+  const urlParams = new URLSearchParams(location.search);
+  let contentFiles;
+
+  if (urlParams.get('local')) {
+    contentFiles = useLocalhostQuickstart(location);
+  } else {
+    contentFiles = usePullRequestQuickstart(location);
+  }
```
aswanson-nr, line 16: @zstix Had a great comment about these hooks. They both do pretty much the same thing, the only difference being the query parameters they use and the function used to grab the files. Because of that, it makes sense to pull this check out of the preview page, combine both hooks, and do the checks there. @josephgregoryii Please let me know what you think about that change!

aswanson-nr (approved):
Tested both PR preview and local preview on the Gatsby preview site, looks good to me!

tabathadelane (approved):
yay!

PR #211, opened 2022-05-24
fix/padding-on-landing-screen

pkosanam:
Description: Observed that the landing page is having issue with the padding

Previous:
image

Now:
image


mdumpati (approved):
approved

PR #224, opened 2022-05-26
NR-15360 applied spinner on "Install New Relic" button

pkosanam:
JIRA ticket: https://issues.newrelic.com/browse/NR-15360

Description: Applied "Spinner" component on the "Install New Relic" button.

When user clicks on the button, user will see a spinner right next to "Install New Relic" and then the application will redirect to a new URL


mdumpati (approved):
APPROVED

mdumpati (approved):
looks good

PR #232, opened 2022-05-31
fix/image-alligned-to-left

pkosanam:
JIRA Ticket: https://issues.newrelic.com/browse/NR-19346

Description:
Observed that few images on the quick start tiles are aligned to the center. Fixed it by aligning all the images to left

Prev:
image

Now:
image


mdumpati (approved):
looks good

PR #249, opened 2022-06-15
NR-20525 Enable linting in PR checks

mickeyryan42:
https://issues.newrelic.com/browse/NR-20525


.github/workflows/run-lint.yml
```diff
@@ -0,0 +1,22 @@
+name: Tests
```
zstix, line 1: We might want to rename this to avoid confusion with the other `Tests` workflow

```suggestion
name: Lint
```

src/components/Overlay.js
```diff
@@ -18,20 +18,23 @@ const Overlay = ({ children, onCloseOverlay, isOpen = false, className }) => {
...
-      onCloseOverlay();
-    }
-  }, [ref.current])
+  const clickListener = useCallback(
+    (e) => {
+      if (ref.current.contains(e.target)) {
+        onCloseOverlay();
+      }
+    },
+    [ref.current]
```
zstix, line 27: Looks like the linter is asking us to add one extra dependency for `useCallback`:

```suggestion
    [ref.current, onCloseOverlay]
```

zstix (commented):
Looks good to me. Two minor comments we might want to address.

I noticed that we're adding a `static/tessen.min.1.14.0.js` file as part of this PR. I thought we already had tessen included in this repository...do you happen to know why this file was added?

.github/workflows/run-lint.yml
mickeyryan42: Thought i did! Good catch

zstix (approved):

PR #253, opened 2022-06-23
NR-21599 Remove margin fixes QuickstartTile width

mickeyryan42:
https://issues.newrelic.com/browse/NR-21599


tabathadelane (approved):
I wasn't able to reproduce the bug. 👍

zstix (approved):

PR #254, opened 2022-06-27
NR-20054 alias components directory in webpack config

mickeyryan42:
https://issues.newrelic.com/browse/NR-20054


josephgregoryii (approved):
Looks awesome! I think we should also have aliases for `hooks/`, `utils`/ and `layouts/` to clean more of the codebase up

We can separate that into different PR's/tickets to make it simpler.

aswanson-nr (approved):

josephgregoryii (approved):
Looks good to me 👍🏻

PR #256, opened 2022-06-29
Source Quickstarts from JSON and Fetch images

zstix:
## Description

This PR changes how we source (load) the quickstarts into GraphQL so that we can "query" for them in Gatsby. Previously were using `gatsby-source-filesystem` and `gatsby-transformer-json` to load the `quickstarts.json` file directly into GraphQL. This was nice because we didn't need to specifically define _all_ the types for each field in the JSON file (the plugins inferred all the types).

Unfortunately, this came with the limitation that it was _very_ hard to update certain fields to be different types. For example, we have `logoUrl` in the JSON file, but we actually want to download that image and turn it into a "File" node for use with sharp. Additionally, there are things that we create (such a `slug`) that don't exist in the JSON file.

This PR introduces a local plugin called `gatsby-source-quickstarts`. This is the recommended way of getting slightly more complex stuff into the GraphQL instance. We need to define the type(s) ourselves, but this now allows us to do the following:

- Download and built with all the logos and screenshots
- Use sharp to optimize all non-SVG images (we can literally query images by the size we want)
- Generate extra properties not in the JSON file without hacks

## Reviewer Notes

Most of the work was done in `plugins/gatsby-source-quickstarts/gatsby-node.js`. This handles bootstrapping (loading the JSON), type definition, fetching images, and resolving extra properties. The rest of the code changes are adjustments to the files that worked with images to use `<GatsbyImage>` (sharp).

I took the time to refactor `<QuickstartImg>` a bit, since we're using it in a few places and the code needed a lot of work to utilize sharp.

There are a few warnings that show up when running this locally about using `childImageSharp` with SVGs. I haven't found a way to suppress this, but it can be ignored, since we're only using that method on non-SVG images.

## Related Ticket(s)
- NR-20529


aswanson-nr (dismissed):
This looks so clean!!! I think we definitely want to do the work to pull in dashboard screenshots as well, we may get less gains on the performance side since they are only loading a few images per details page, but we get the confidence in knowing that the images are stored in the CDN with the rest of the site and changes to the quickstart repo can't affect the live site.

plugins/gatsby-source-quickstarts/gatsby-node.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+};
+
+/**
+ * Rather than inferring the type from the JSON object, we specifically
+ * define the types here. This allows us to return different values for
+ * some of the fields (in `createResolvers`) and to fetch remote resources
+ * as needed (in `sourceNodes`).
+ *
+ * @todo Verify that `relatedResources` gets properly added from theme.
```
zstix, line 41:
```suggestion
```

plugins/gatsby-source-quickstarts/gatsby-node.js
```diff
@@ -0,0 +1,239 @@
...
+      summary: String
+      keywords: [String]
+      authors: [String]
+      logo: File
+      level: QuickstartSupportLevel
+      documentation: [QuickstartDocumentation]
+      alerts: [QuickstartAlertCondition]
+      dashboards: [QuickstartDashboard]
+      installPlans: [QuickstartInstallPlan]
+
```
zstix, line 59:
```suggestion
```

src/pages/index.jsx
```diff
@@ -757,37 +756,24 @@ export const pageQuery = graphql`
...
-        summary
-        installPlans {
-          id
-          name
+        logo {
+          ext
+          publicURL
+          childImageSharp {
+            gatsbyImageData(
+              height: 45
```
zstix, line 772: TODO: add a comment calling out why we have `45`(px) as a hard-coded value.

src/templates/QuickstartDetails.js
```diff
@@ -196,15 +197,32 @@ export const pageQuery = graphql`
...
+      }
       packUrl
       dashboards {
         description
         name
-        screenshots
+        screenshots {
+          childImageSharp {
+            gatsbyImageData(
+              height: 280
```
zstix, line 220: This is really compromising the quality of the screenshots. Let's let CSS size the image and bring them in at full size. WEBP is still going to give us some filesize improvements (~50%)
```suggestion
```

src/components/QuickstartDashboards.js
```diff
@@ -116,16 +118,16 @@ const QuickstartDashboards = ({ quickstart }) => {
                       `}
                     >
                       <a
-                        href={imgUrl}
+                        href={node.publicURL}
```
zstix, line 121: It looks like this `href` is not getting set. It's possible we might not have `publicURL`.

src/templates/QuickstartDetails.js
```diff
@@ -196,15 +197,32 @@ export const pageQuery = graphql`
...
+          )
+        }
+      }
       packUrl
       dashboards {
         description
         name
-        screenshots
+        screenshots {
+          childImageSharp {
```
zstix, line 218:
```suggestion
        screenshots {
          publidURL
          childImageSharp {
```

mickeyryan42 (dismissed):
Reviewed from a purely code perspective since I'm not super knowledgeable with the Gatsby side.  Looks awesome!

aswanson-nr (changes requested):
Looking pretty much good to go to me! Images now link out to the original in a new tab.
The one issue I noticed is that the dashboard images are being loaded as full width while maintaining their height, see the attached screenshot.

### Live site
Screen Shot 2022-07-05 at 1 55 30 PM

### PR preview of this pull request
Screen Shot 2022-07-05 at 1 52 31 PM

aswanson-nr (approved):

PR #259, opened 2022-06-30
Changes to page pathname close mobile menu

tabathadelane:
The header now detects when a user clicks the back button while the mobile menu overlay is open and closes it while navigating.

I chose to use the logic in the gatsby-theme-newrelic [MobileNavigation](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/MobileNavigation.js) component. I was having a lot of styling issues come up when trying to import the component directly. 

I was seeing odd behavior in the local server where sometimes when clicking the hamburger icon, I would have to click it twice to get the overlay to open. I can't seem to reproduce it in a build, so please check this when reviewing.

### DIFF
--- a/src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
+++ b/src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
 import AnnouncementBanner from '@newrelic/gatsby-theme-newrelic/src/components/AnnouncementBanner';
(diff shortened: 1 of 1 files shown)


aswanson-nr (dismissed):
A wonderfully simple solution!!

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
```diff
@@ -85,10 +94,11 @@ const GlobalHeader = ({ className, activeSite }) => {
     }
   }, [isOpen]);
 
-  const hideLogoText = useMedia({ maxWidth: '350px' });
-  const location = useLocation();
-  const UserIsInMainPage = location.pathname === '/instant-observability/';
-  const showGetStarted = !!UserIsInMainPage;
+  React.useEffect(() => {
```
mickeyryan42, line 97: I can't remember where we left the discussion, do we want to import the destructured function vs using it off the React object? Just a side note, but otherwise i cant reproduce the bug 👍

mickeyryan42 (approved):

mickeyryan42 (approved):

josephgregoryii (approved):
:shrek: :shipit:

PR #260, opened 2022-07-01
NR-21565 Fix overlapping banner and categories input

mickeyryan42:
https://issues.newrelic.com/browse/NR-21565

I'm not sure if I'm a big fan of this solution, but it seems to fix the issue.  Putting it up for discussion and ideas.


josephgregoryii (approved):
As we talked about in standup, I'm not seeing a fairly quick/easy way to improve this based from looking at other breakpoints/responsiveness. I think it looks great for now 👍🏻

tabathadelane (approved):
I was unable to reproduce it in safari or chrome 👍

PR #263, opened 2022-07-06
Sort quickstart components by available content

josephgregoryii:
## Summary

This PR serves the purpose of rendering the component sections of the Landing Page with content _before_ the sections without content.

> NOTE: Currently, the `Dashboards.jsx` file has the "What's included?" header inside of the component. There will be another PR to address that change. Example of header issues below.

#### Example

Screen Shot 2022-07-06 at 1 14 56 PM

### Considerations
- As mentioned above, there will be another PR for React/CSS changes. That PR will be merged into this `feat` branch so that both changes are merged to `main` at the same time.

### New Expected Behavior:
1. `Dashboards(0)`, `Alerts(0)`, `Data Sources(1)` -> `Data Sources(1)`, `Dashboards(0)`, `Alerts(0)`
2. `Dashboards(1)`, `Alerts(2)`, `Data Sources(1)` ->  `Dashboards(1)`, `Alerts(2)`, `Data Sources(1)`
3. `Dashboards(0)`, `Alerts(2)`, `Data Sources(1)` ->  `Alerts(2)`, `Data Sources(1)`, `Dashboards(0)`,

### DIFF
--- a/src/templates/QuickstartDetails.js
+++ b/src/templates/QuickstartDetails.js
@@ -27,22 +27,28 @@ const layoutContentSpacing = css`
 `;
 
 const moveToBack = (componentsAndCounts) => {
+  const n = componentsAndCounts.length;
   const newOrder = [];
 
-  // traverse array from back to front
-  // to remain same order of components for render.
-  let i = componentsAndCounts.length - 1;
-
-  while (i >= 0) {
-    if (componentsAndCounts[i].count === 0) {
+  // add components with content
+  // to the beginning of array
+  let i = 0;
+  while (i < n) {
+    if (componentsAndCounts[i].count !== 0) {
       // Push element to end of array.
       // This keeps the same order if all components
       // have a length of 0.
-      newOrder.push(componentsAndCounts[i--]);
-    } else {
-      // if there are components, add to the front of the array
-      newOrder.unshift(componentsAndCounts[i--]);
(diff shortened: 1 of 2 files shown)


src/templates/QuickstartDetails.js
```diff
@@ -26,6 +26,62 @@ const layoutContentSpacing = css`
...
+  // Check to see if we need to move empty components to
+  // the end of the array
+  const hasEmptyComponent = componentsAndCounts.filter(
+    (component) => component.count
+  );
+
+  return hasEmptyComponent.length > 0
+    ? moveToBack(componentsAndCounts)
+    : componentsAndCounts;
+};
```
aswanson-nr, line 83: I could be wrong here, but I think we could replace all this with
```suggestion
  // Check to see if we need to move empty components to
  // the end of the array
  return componentsAndCounts.sort((a, b) => {
    if (a.length < 1) {
      return -1;
    } else if (b.length < 1) {
      return 1;
    } else {
      return 0;
    }
  })
};
```

josephgregoryii: Yup! i was definitely missing something. Only correct to this is 

```jsx
if (a.count < 1) {
      return 1;
    } else if (b.count < 1) {
      return -1;
```
in order to keep it in the correct order 👍🏻

aswanson-nr (approved):

mickeyryan42 (approved):

PR #265, opened 2022-07-06
React/CSS updates for content organization

josephgregoryii:


zstix (dismissed):

src/components/WhatsIncluded/WhatsIncludedHeader.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { css } from '@emotion/react';
+import Share from '@components/Share';
+import { useLocation } from '@react/router';
```
aswanson-nr, line 4: This should be 
```suggestion
import { useLocation } from '@reach/router';
```

src/templates/QuickstartDetails.js
```diff
@@ -96,9 +98,15 @@ const QuickstartDetails = ({ data, location }) => {
             ${layoutContentSpacing};
             > * {
               padding-top: 3rem;
+              @media screen and (min-width: ${QUICKSTARTS_COLLAPSE_BREAKPOINT}) {
+                :nth-child(2) {
+                  padding-top: 1rem;
+                }
```
aswanson-nr, line 104: Instead of using `:nth-child` here, we could just apply this to the `<WhatsIncludedHeader />` element and use `padding-bottom`. If you want to keep using `:nth-child`, lets use `:nth-of-type` instead. This is due to some SSR behavior with emotion, more details [here](https://emotion.sh/docs/ssr#default-approach)

aswanson-nr (commented):
I had one last change! But overall this looks good to go!

aswanson-nr (approved):
Looks good to go!

mickeyryan42 (approved):

mickeyryan42 (approved):

PR #268, opened 2022-07-11
Align catalog headers and carousel margins

tabathadelane:
this aligns the edges of the carousel slides with its headers and also with the sidebar headers

Screen Shot 2022-07-11 at 10 44 50 AM

### DIFF
--- a/src/components/QuickstartDashboards.js
+++ b/src/components/QuickstartDashboards.js
@@ -25,9 +25,9 @@ const QuickstartDashboards = ({ quickstart }) => {
         css={css`
           width: 62px;
           height: 62px;
-          margin-left: -3.5rem;
+          margin-left: -2rem;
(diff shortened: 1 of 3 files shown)


josephgregoryii (approved):

aswanson-nr (approved):
This looks so good!!! Everything feels so clean and lined up now.
I don't like the idea of adding more `!important` labels, but that seems like the only way to style this slider.

tabathadelane:
> This looks so good!!! Everything feels so clean and lined up now. I don't like the idea of adding more `!important` labels, but that seems like the only way to style this slider.

Thanks. It is pretty satisfying! And yeah, i don't love the !important either. I tried to avoid it, but nothing was working. Then I saw we did that the last time we edited it, so It's probably the only way.

zstix (approved):

PR #271, opened 2022-07-12
Fix SVG errors during development & builds

zstix:
## Description

This change, while somewhat clunky in the query, prevents the following error from showing up during builds and local development:

```
warn You can't use childImageSharp together with logo.svg — use publicURL instead. The childImageSharp portion of the query in this file will return null:
```

The reason for this error is that sharp does not work with SVG files - they are already performant and do not need any additional optimizations. We needed a separate field so that we could avoid
querying `childImageSharp`.

## Reviewer Notes

I'm curious to hear if folks find this change to be worth it to get rid of these warnings.

### DIFF
--- a/plugins/gatsby-source-quickstarts/gatsby-node.js
+++ b/plugins/gatsby-source-quickstarts/gatsby-node.js
@@ -50,6 +50,7 @@ exports.createSchemaCustomization = ({ actions }) => {
       summary: String
       keywords: [String]
       authors: [String]
+      logoSvg: File
       logo: File
       level: QuickstartSupportLevel
       documentation: [QuickstartDocumentation]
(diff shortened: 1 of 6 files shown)


plugins/gatsby-source-quickstarts/gatsby-node.js
```diff
@@ -173,7 +176,8 @@ exports.sourceNodes = async ({
         documentation: quickstart.documentation,
         alerts: quickstart.alerts,
         installPlans: quickstart.installPlans,
-        logo: logoNode || null,
+        logo: !isLogoSvg ? logoNode : null,
+        logoSvg: isLogoSvg ? logoNode : null,
```
aswanson-nr, line 180: It might be nice to add a comment here as to why we are splitting out the SVGs from the rest of the images.

PR #272, opened 2022-07-12
Sort and filter quickstart hooks

josephgregoryii:


src/hooks/useFilteredQuickstarts.jsx
```diff
@@ -0,0 +1,136 @@
...
+ * @param {(Function) => Array} Callback function that filters quickstart array
+ */
+const filterQuickstarts = (quickstarts) => (keyword) =>
+  quickstarts.filter((product) => product.keywords.includes(keyword));
+
+/**
+ * Custom hook to get filtered quickstarts
+ * @param {Array} array of quickstarts
+ */
+const useFilteredQuickstarts = (quickstarts, search, category) => {
```
aswanson-nr, line 23: This is more to invoke a discussion, than to say what it should be one way or another.
Does this even need to be a hook? If the search/filter hook triggers a rerender this then that will run all calls within the functional component (the index page). So if we just have these run on the list of quickstarts, they will be called again each render.

src/pages/index.jsx
```diff
@@ -403,41 +274,46 @@ const QuickstartsPage = ({ data, location }) => {
...
+              {!loadComplete && <Spinner />}
+              {loadComplete &&
+                categoriesWithCount.map(({ displayName, value, count }) => (
+                  <Button
+                    type="button"
+                    key={value}
+                    disabled={count === 0}
+                    variant={Button.VARIANT.PRIMARY}
+                    onClick={() => {
+                      handleParams('category', 'search')(value, search);
```
aswanson-nr, line 287: To clear up what this is doing we could declare a variable at the top of the render block for this.
So at the top of the render block:
```javascript
const handleSearchAndCategory = handleParams('category', 'search');
```

and down here we could call it:
```javascript
onClick={()=> handleSearchAndCategory(value, search)}
```

src/hooks/useSearchAndCategory.jsx
```diff
@@ -0,0 +1,74 @@
+import { useState, useEffect } from 'react';
+import { navigate } from 'gatsby';
+import { useTessen } from '@newrelic/gatsby-theme-newrelic';
+import CATEGORIES from '@data/instant-observability-categories';
+
+const isValueValid = (value) => value !== null && value !== undefined;
```
aswanson-nr, line 6: Since this is just used for URL search parameter values and we are only using it locally, I think we could simplify the name to `isValid`.
```suggestion
const isValid = (value) => value !== null && value !== undefined;
```

src/hooks/useSearchAndCategory.jsx
```diff
@@ -0,0 +1,74 @@
...
+  const tessen = useTessen();
+
+  useEffect(() => {
+    const params = new URLSearchParams(location.search);
+    const searchParam = params.get('search');
+    const categoryParam = params.get('category');
+    const validCategory = CATEGORIES.some((cat) => cat.value === categoryParam);
+
+    setSearch(searchParam || '');
+    setCategory(categoryParam && validCategory ? categoryParam : '');
```
aswanson-nr, line 21: We could clarify this by creating a variable, `const isCategoryValid = categoryParam && validCategory;`
Then when we use it here, it's super clear what we're doing.
```javascript
setCategory(isCategoryValid ? categoryParam : '');
```

src/pages/index.jsx
```diff
@@ -293,7 +163,7 @@ const QuickstartsPage = ({ data, location }) => {
   const renderGoToTopButton = () => {
     return (
       <Button
-        onClick={topFunction}
+        onClick={() => topFunction()}
```
aswanson-nr, line 166: I'm curious about your reasoning for this change, because it _should_ work fine when the `topFunction` is passed in without the wrapping arrow function.

src/utils/getFilteredQuickstarts.js
```diff
@@ -0,0 +1,128 @@
...
+const shiftCodestream = (alphaSortedQuickstarts, codestreamIndex) => {
+  // Hard-code for moving codestream object to front of sortedQuickstarts array
+  // hardcoded codestream uuid
+  const codestreamObject = alphaSortedQuickstarts[codestreamIndex];
+  return [
+    codestreamObject,
+    ...alphaSortedQuickstarts.slice(0, codestreamIndex),
+    ...alphaSortedQuickstarts.slice(codestreamIndex + 1),
+  ];
+};
```
aswanson-nr, line 81: This is totally optional, but we could simplify this moving the codestream quickstart to the front with a `sort`.
```javascript
const codestreamId = '29bd9a4a-1c19-4219-9694-0942f6411ce7';
const quickstarts = alphaSortedQuickstarts.sort((a, b) => {
  if (a.id === codestreamId) {
    return 1;
  }
  if (b.id === codestreamId) {
    return 1;
  }
  return 0;
}
```

src/utils/getFilteredQuickstarts.js
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Custom hook to get filtered quickstarts
+ * @param {Array} array of quickstarts
+ */
+const getFilteredQuickstarts = (quickstarts, search, category) => {
```
aswanson-nr, line 87: Another totally optional change, for me the `get` part of the name symbolizes that it retrieves the quickstarts, when in reality it is taking in the quickstarts and filtering them in various ways. We could possibly change it to `filterQuickstarts` since that's all it is doing.

src/pages/index.jsx
josephgregoryii: It was only to remove a pesky eslint warning/error. This was only to remove it. If you would like me to change it back, happy to!

src/pages/index.jsx
```diff
@@ -324,12 +194,13 @@ const QuickstartsPage = ({ data, location }) => {
         location={location}
         type="quickstarts"
       />
-      <IOBanner
-        search={search}
-        setSearch={setSearch}
-        setIsSearchInputEmpty={setIsSearchInputEmpty}
-        handleSearch={handleSearch}
-      />
+      {handleParam && (
```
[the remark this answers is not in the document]
josephgregoryii: Ah, i was getting some weird error earlier but now i'm not getting it - consider it removed!

src/pages/index.jsx
josephgregoryii: Great callout! resolved

src/hooks/useSearchAndCategory.jsx
josephgregoryii: done

src/hooks/useSearchAndCategory.jsx
josephgregoryii: great catch

aswanson-nr (approved):

zstix (approved):

PR #276, opened 2022-07-14
Refactor index page

tabathadelane:
So far we have

- extracted the slick carousel settings into a constants file. 

Do we need to do more?

### DIFF
--- a/src/pages/index.jsx
+++ b/src/pages/index.jsx
@@ -12,19 +12,20 @@ import IOBanner from '@components/IOBanner';
 import IOSeo from '@components/IOSeo';
 import Overlay from '@components/Overlay';
 import PropTypes from 'prop-types';
-import { QUICKSTARTS_COLLAPSE_BREAKPOINT } from '@data/constants';
+import {
+  QUICKSTARTS_COLLAPSE_BREAKPOINT,
+  TRIPLE_COLUMN_BREAKPOINT,
+  DOUBLE_COLUMN_BREAKPOINT,
+  SINGLE_COLUMN_BREAKPOINT,
+} from '@data/constants';
+import { indexSettings } from '../data/slick-settings';
 import QuickstartTile from '@components/QuickstartTile';
 import Slider from 'react-slick';
 import SuperTiles from '@components/SuperTiles';
 import { css } from '@emotion/react';
 import { graphql } from 'gatsby';
-import LeftArrowSVG from '@components/Icons/LeftArrowSVG';
-import RightArrowSVG from '@components/Icons/RightArrowSVG';
 import featherIcons from '../@newrelic/gatsby-theme-newrelic/icons/feather';
 
-const TRIPLE_COLUMN_BREAKPOINT = '1420px';
-const DOUBLE_COLUMN_BREAKPOINT = '1180px';
-const SINGLE_COLUMN_BREAKPOINT = '900px';
 const COLUMN_BREAKPOINT = '1131px';
 // used to set the height of the Spinner to reduce layout shift on page load
 const TILE_HEIGHT = '362px';
@@ -72,70 +73,6 @@ const QuickstartsPage = ({ data, location }) => {
     if (!found || !found.value) return defaultName;
 
     return found.displayName;
-  };
-
-  // Settings for Slick-Carousel
-  const settings = {
-    dots: false,
-    infinite: false,
-    speed: 300,
-    slidesToShow: 4,
-    slidesToScroll: 4,
-    adaptiveHeight: false,
-    adaptiveWidth: true,
-    mobileFirst: true, // necessary for breakpoints to work as expected
-    prevArrow: (
-      <button type="button">
-        <LeftArrowSVG
-          className="slick-prev"
-          css={css`
-            width: auto;
-            height: auto;
-            margin: 0 1.5rem;
-          `}
-        />
-      </button>
-    ),
-    nextArrow: (
-      <button type="button">
-        <RightArrowSVG
-          className="slick-next"
-          css={css`
-            width: auto;
-            height: auto;
-            margin: 0 1.5rem;
-          `}
-        />
-      </button>
-    ),
-
-    responsive: [
-      {
-        breakpoint: parseInt(TRIPLE_COLUMN_BREAKPOINT),
-        settings: {
-          slidesToShow: 3,
-          slidesToScroll: 3,
-          infinite: true,
-          dots: false,
-        },
-      },
-      {
-        breakpoint: parseInt(DOUBLE_COLUMN_BREAKPOINT),
-        settings: {
-          slidesToShow: 2,
-          slidesToScroll: 2,
-          infinite: true,
-          dots: false,
-        },
-      },
-      {
-        breakpoint: parseInt(SINGLE_COLUMN_BREAKPOINT),
-        settings: {
-          slidesToShow: 1,
-          slidesToScroll: 1,
-        },
-      },
-    ],
   };
 
   const handleScroll = () => {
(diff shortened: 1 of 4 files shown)


josephgregoryii (dismissed):
I think this is a solid start for now. I'm happy to merge this PR and discuss next steps in standup.

I think we have two ways to further improve
1) Create simple components that we can bring into the index page that will have majority of the CSS to further lighten the load of the index page
2) Create styled components in a subdirectory in `components` to accomplish the same feat.

mickeyryan42:
I also have a branch that moved the category aside and mobile category dropdown into separate components. If thats useful we can look at merging to this branch.  I think it simplified the file by a few hundred lines.

src/pages/index.jsx
```diff
@@ -253,70 +191,32 @@ const QuickstartsPage = ({ data, location }) => {
...
+            search={search}
+          />
         </aside>
         <div
           css={css`
             grid-area: main;
             padding: 1.5rem;
           `}
         >
+<<<<<<< HEAD
```
josephgregoryii, line 208: Looks like we have some pushed merge code lines

zstix:
```
pages/
  |- index.jsx
components/
  |- foo.jsx
  |- indexComponents/
  |- |- aside.jsx
```

src/components/indexComponents/CategoryDropdown.js
```diff
@@ -0,0 +1,164 @@
...
+import { QUICKSTARTS_COLLAPSE_BREAKPOINT } from '@data/constants';
+
+const CategoryDropdown = ({
+  loadComplete,
+  category,
+  categoriesWithCount,
+  setIsCategoriesOverlayOpen,
+  isCategoriesOverlayOpen,
+  closeCategoriesOverlay,
+  getDisplayName,
```
aswanson-nr, line 16: We can probably pull this function out to a helper file, that way we can import into each component instead of passing it in as a prop.

src/components/indexComponents/CategoryDropdown.js
```diff
@@ -0,0 +1,164 @@
...
+
+import { QUICKSTARTS_COLLAPSE_BREAKPOINT } from '@data/constants';
+
+const CategoryDropdown = ({
+  loadComplete,
+  category,
+  categoriesWithCount,
+  setIsCategoriesOverlayOpen,
+  isCategoriesOverlayOpen,
+  closeCategoriesOverlay,
```
aswanson-nr, line 15: I think we could pull this state into the component, instead of passing it through from the index page.

src/components/indexComponents/CategoryDropdown.js
```diff
@@ -0,0 +1,164 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { css } from '@emotion/react';
+import { Button, Icon, Spinner } from '@newrelic/gatsby-theme-newrelic';
+import Overlay from '@components/Overlay';
+
+import { QUICKSTARTS_COLLAPSE_BREAKPOINT } from '@data/constants';
+
+const CategoryDropdown = ({
+  loadComplete,
```
aswanson-nr, line 10: This would be a change in functionality, so it's definitely not a required change. However, I can't think of any reason this should need to wait for the page to hit the browser instead of rendering it at build time.

src/components/indexComponents/CategoryList.js
```diff
@@ -0,0 +1,119 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { css } from '@emotion/react';
+import { Button, Spinner } from '@newrelic/gatsby-theme-newrelic';
+import { QUICKSTARTS_COLLAPSE_BREAKPOINT } from '@data/constants';
+
+const CategoryList = ({
+  loadComplete,
```
aswanson-nr, line 8: Same comment as the other category component

aswanson-nr (commented):
I saw a few spots we could clean up, but overall looks fantastic. None of these comments should stop the merge.

src/components/indexComponents/CategoryDropdown.js
mickeyryan42: Yeah thats a great point, it cleans up the prop drilling by 3 props too.

src/components/indexComponents/CategoryDropdown.js
mickeyryan42: ill experiment and remove it and see if i notice any differences!

aswanson-nr: Ah, I think I know why this was here in the first place. If a user loads the page with a category query parameter, you get a flash of the `All` category before it updates to the correct category. This was in there to have it wait to render until it know which category is selected. User experience wise, I think we should keep this in. Sorry for the flip flop!

src/components/indexComponents/CategoryDropdown.js
```diff
@@ -0,0 +1,163 @@
...
+import Overlay from '@components/Overlay';
+
+import getDisplayName from '@utils/getDisplayName';
+import { QUICKSTARTS_COLLAPSE_BREAKPOINT } from '@data/constants';
+
+const CategoryDropdown = ({
+  category,
+  categoriesWithCount,
+  handleParam,
+  loadComplete,
```
aswanson-nr, line 14: Okay, I'm so sorry about another comment on this. But I was thinking, all this hook is doing is changing from `false` to `true` when it hits the browser, so we might as well get rid of the prop drilling and manage that state within this component. (Same with other category component)

src/components/indexComponents/CategoryDropdown.js
```diff
@@ -0,0 +1,162 @@
...
+import getDisplayName from '@utils/getDisplayName';
+import { QUICKSTARTS_COLLAPSE_BREAKPOINT } from '@data/constants';
+
+const CategoryDropdown = ({ category, categoriesWithCount, handleParam }) => {
+  const [isCategoriesOverlayOpen, setIsCategoriesOverlayOpen] = useState(false);
+  const [loadComplete, setLoadComplete] = useState(false);
+
+  useEffect(() => {
+    setLoadComplete(true);
+  }, [categoriesWithCount]);
```
aswanson-nr, line 16: I don't think we need this to be a dependency for this `useEffect`. Since we really just want it to trigger once when the page loads.
```suggestion
  }, []);
```

src/components/indexComponents/CategoryList.js
```diff
@@ -0,0 +1,123 @@
...
+  category,
+  categoriesWithCount,
+  handleSearchAndCategory,
+  search,
+}) => {
+  const [loadComplete, setLoadComplete] = useState(false);
+
+  useEffect(() => {
+    setLoadComplete(true);
+  }, [categoriesWithCount]);
```
aswanson-nr, line 17:
```suggestion
  }, []);
```

src/pages/index.jsx
```diff
@@ -47,97 +53,12 @@ const QuickstartsPage = ({ data, location }) => {
...
-  const [isCategoriesOverlayOpen, setIsCategoriesOverlayOpen] = useState(false);
-  // variable to check if the page load completed
   const [loadComplete, setLoadComplete] = useState(false);
 
-  // mark the value as true, if the page is loaded
   useEffect(() => {
-    if (categoriesWithCount) {
       setLoadComplete(true);
-    }
   }, [categoriesWithCount]);
```
aswanson-nr, line 60: Same here as well
```suggestion
  }, []);
```

aswanson-nr (commented):
I don't think we need to the `useEffect` to have a dependency on the categories object, other wise looks good to go!

src/components/indexComponents/CategoryDropdown.js
mickeyryan42: ahh good point. removed them

PR #277, opened 2022-07-18
Shadow 404 page

aswanson-nr:
# Summary
Pull the 404 page into the site from the theme and reduce functionality to work with IO.

**Note**: The preview does not seem to work well with the 404 page, so you'll need to pull it down and `yarn build && yarn serve`

## Screenshot
Screen Shot 2022-07-18 at 3 12 03 PM

### DIFF
--- a/src/@newrelic/gatsby-theme-newrelic/pages/404.js
+++ b/src/@newrelic/gatsby-theme-newrelic/pages/404.js
@@ -12,6 +12,8 @@ import {
 
 import QuickstartLayout from '../../../layouts/QuickStartLayout';
 
+const uuidRegex = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
+
 const NotFoundPage = ({ location, pageContext: { swiftypeEngineKey } }) => {
   const [searchTerm, setSearchTerm] = useState(null);
   const [searchResult, setSearchResult] = useState(null);
@@ -60,7 +62,13 @@ const NotFoundPage = ({ location, pageContext: { swiftypeEngineKey } }) => {
   }, [searchTerm, swiftypeEngineKey]);
 
   useEffect(() => {
-    setSearchTerm(location.pathname.split('/').join(' '));
+    setSearchTerm(
+      location.pathname
+        .split('/')
+        .filter((p) => p !== 'instant-observability')
+        .filter((p) => !uuidRegex.test(p))
+        .join(' ')
+    );
   }, [location.pathname]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/@newrelic/gatsby-theme-newrelic/pages/404.js
```diff
@@ -0,0 +1,207 @@
...
+    origin: PropTypes.string.isRequired,
+    pathname: PropTypes.string.isRequired,
+  }).isRequired,
+  pageContext: PropTypes.shape({
+    themeOptions: PropTypes.object.isRequired,
+    swiftypeEngineKey: PropTypes.string.isRequired,
+  }).isRequired,
+};
+
+const displaySearchResults = (term, results) => {
```
zstix, line 156: We might want to consider moving this to it's own file. Additionally, we should probably define the `propTypes` for this component so that our linter doesn't get angry at us.

zstix (commented):
It looks like we get better results without the `instant-observability` part of the url. We should consider only using _part_ of the URL (without UUID and the path prefix) to populate the search.

zstix (dismissed):
Looks good to me! Thanks for making those updates.

src/@newrelic/gatsby-theme-newrelic/pages/404.js
```diff
@@ -0,0 +1,227 @@
...
+const uuidRegex = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
+
+const NotFoundPage = ({ location, pageContext: { swiftypeEngineKey } }) => {
+  const [searchTerm, setSearchTerm] = useState(null);
+  const [searchResult, setSearchResult] = useState(null);
+  const tessen = useTessen();
+
+  const getSearchResults = useCallback(async () => {
+    const search = async () => {
+      const res = await fetch(
```
mickeyryan42, line 24: Should this be in a try/catch?

src/@newrelic/gatsby-theme-newrelic/pages/404.js
```diff
@@ -0,0 +1,239 @@
...
+      throw new Error(
+        `Could not fetch related pages. ${res.status} ${res.statusText}`
+      );
+    }
+
+    const { records } = await res.json();
+
+    return records.page;
+  } catch (err) {
+    console.error(err);
```
aswanson-nr, line 233: I'm don't think we want to just log this to the console. Perhaps we could send a tessen event?

src/@newrelic/gatsby-theme-newrelic/pages/404.js
aswanson-nr: Great point @mickeyryan42 ! I got the PR updated to include a try/catch block

mickeyryan42 (approved):
Sorry for forgetting to review before signing off!

zstix (approved):

PR #278, opened 2022-07-18
Landing page cleanup and gql fragments

josephgregoryii:
### Summary
This PR handles cleaning up `QuickstartDetails` file. Along with some function extractions, the main code change is to utilize `fragments` through GraphQL. 

#### Considerations
- I went with the approach to include  fields for the `pageQuery` at each fragment to be explicit with what each component needs for data. This leads to some minimal duplication when also looking at grandchildren components. I'm curious what others think of this approach, and if we do not need it to be as verbose.
- Following the naming convention from UI Unconf talk, I followed the naming convention to be `ReactComponent_propName`, where `ReactComponent` specifies the name of the component that is fetching the data, and `propName` to indicate where the data is getting retrieved from
- Since we are hosted by Gatsby, I followed their [docs](https://www.gatsbyjs.com/docs/reference/graphql-data-layer/using-graphql-fragments/) to achieve fragments. String interpolation isn't allowed using `graphql` from Gatsby, so using `${Component.fragments.quickstart}` would not work.

### DIFF
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -225,9 +225,6 @@ InstallButton.propTypes = {
 
 export const fragmentQuery = graphql`
   fragment InstallButton_quickstart on Quickstarts {
-    id
-    name
-    packUrl
     installPlans {
       id
       name
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
+import { graphql } from 'gatsby';
 import { Button, Link, useTessen } from '@newrelic/gatsby-theme-newrelic';
 import {
   getPackNr1Url,
@@ -221,5 +222,22 @@ InstallButton.propTypes = {
   location: PropTypes.object.isRequired,
   buttonStyle: PropTypes.oneOf(['PRIMARY', 'SECONDARY']),
 };
+
+export const fragmentQuery = graphql`
+  fragment InstallButton_quickstart on Quickstarts {
+    id
+    name
+    packUrl
+    installPlans {
+      id
+      name
+    }
+    documentation {
+      name
+      url
+      description
+    }
+  }
+`;
 
 export default InstallButton;
--- a/src/components/QuickstartHowToUse.js
+++ b/src/components/QuickstartHowToUse.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { graphql } from 'gatsby';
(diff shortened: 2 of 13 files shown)


src/components/InstallButton.js
```diff
@@ -222,4 +223,21 @@ InstallButton.propTypes = {
   buttonStyle: PropTypes.oneOf(['PRIMARY', 'SECONDARY']),
 };
 
+export const fragmentQuery = graphql`
+  fragment InstallButton_quickstart on Quickstarts {
```
josephgregoryii, line 227: Naming convention: 
`ReactComponent_PropName`

src/components/QuickstartHowToUse.js
```diff
@@ -124,4 +125,10 @@ QuickstartHowToUse.propTypes = {
   layoutContentSpacing: PropTypes.object,
 };
 
+export const fragmentQuery = graphql`
+  fragment QuickstartHowToUse_quickstart on Quickstarts {
+    ...InstallButton_quickstart
+  }
+`;
```
zstix, line 132: Do we need to export a fragment that calls another fragment? This feels a bit like "prop-drilling" but for fragments.

I wonder if it would be possible to just reference `InstallButton_quickstart` where we are doing the top-level query, rather than referencing `QuickstartHowToUse_quickstart`.

src/components/WhatsIncluded/Alerts.js
```diff
@@ -55,4 +56,12 @@ Alerts.propTypes = {
   quickstart,
 };
 
+export const fragmentQuery = graphql`
+  fragment Alerts_quickstart on Quickstarts {
+    title
+    packUrl
```
zstix, line 62: I noticed that we're querying for a lot of the same fields in these different fragments. Do we know if the final query that Gatsby pulls together removes those duplicates? Otherwise this might have a negative impact on our builds / bundles.

zstix: The bundle seems to be the same size, so this is probably a non-issue 👍

src/components/QuickstartHowToUse.js
josephgregoryii: Seeing how Gatsby likes to implement fragments, i think this is a great callout. I'll get that sorted

src/components/WhatsIncluded/Alerts.js
josephgregoryii: I _think_ graphql handles duplicate fields on the backend. I tried it using `graphiql` and there didnt seem to be any issues, however I think its best to remove the duplicates anyway 👍🏻 

updated

aswanson-nr (approved):

PR #281, opened 2022-07-19
NR-21567: Alignment for quickstart arrow icon on safari

mickeyryan42:
https://issues.newrelic.com/browse/NR-21567


tabathadelane (approved):
I looks good on my end, but I do could only try to reproduce in safari dev tools. 👍

josephgregoryii (approved):

PR #292, opened 2022-07-25
Add workflow to clear cache on push

aswanson-nr:
## Summary
Adds a workflow to clear the website's fastly cache whenever we push new code. This workflow can also be triggered via a workflow run.

### DIFF
--- a/.github/workflows/clear-website-cache.yml
+++ b/.github/workflows/clear-website-cache.yml
@@ -1,0 +1,23 @@
+name: Clear website cache
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+
+jobs:
+  clear-cache:
+    runs-on: ubuntu-latest
+    env:
+      TOKEN: ${{ secrets.WEBSITE_CACHE_TOKEN }}
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/clear-website-cache.yml
```diff
@@ -0,0 +1,23 @@
+name: Clear website cache
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
josephgregoryii, line 7: I may be misinterpreting this - why would we not want it to run once a PR is merged? I know we have branch protection, however in the chance that branch protection is turned _off_ and mistakenly wasn't turned back on, this would be triggered from that push to main.

Not really too worried about that, just an edge case i'm thinking of.

josephgregoryii (approved):

.github/workflows/clear-website-cache.yml
aswanson-nr: That's a good point, but considering any push to main is deployed to Gatsby Cloud, wouldn't we want to clear the cache as well?

zstix (approved):

PR #295, opened 2022-07-26
Dashboard images no longer stretch on safari

tabathadelane:
I looked for this bug in multiple page widths and in Safari, Chrome, and Firefox on a local build + serve. It appears to be gone, while keeping the desired behavior of the images across breakpoints. Please double check.

### DIFF
--- a/src/components/QuickstartDashboards.js
+++ b/src/components/QuickstartDashboards.js
@@ -88,10 +88,8 @@ const QuickstartDashboards = ({ quickstart }) => {
                         src={node.publicURL}
                         alt={`${dashboard.name} screenshot ${index}`}
                         css={css`
-                          width: 100%;
                           height: 17.5rem;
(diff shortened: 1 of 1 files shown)


tabathadelane:
I noticed the jumping/centering, but really only when I was resizing the browser. I don't feel like that is a super common user behavior if the website is responsive. It's worth talking about though for sure.

mickeyryan42:
> I noticed the jumping/centering, but really only when I was resizing the browser. I don't feel like that is a super common user behavior if the website is responsive. It's worth talking about though for sure.

Yeah totally agree. I saw the same, mostly on resizing. It was hard to tell if it was due to the resizing or only on those random screen sizes. Either way, definitely not anything pressing

aswanson-nr (approved):
yeah the jumping/jittery behavior exists on all browsers currently, we just couldn't see it on Safari because one image was taking up the whole carousel. This looks good to me!

PR #298, opened 2022-07-28
Selectively import dependencies from the gatsby theme

aswanson-nr:
# Summary
* Import components, hooks, and functions directly from the theme,
  instead of pulling them from the `export` index file.
* This reduces our webpack bundle size by only including dependencies
  from the theme that we actually need

### Before
Screen Shot 2022-07-28 at 1 13 21 PM-min
Screen Shot 2022-07-28 at 1 13 32 PM-min

### After
Screen Shot 2022-07-28 at 1 15 35 PM-min

Screen Shot 2022-07-28 at 1 15 43 PM-min

### Savings
This reduces the overall total size of our bundles by about 75% and cuts our core app bundle by 50%


mickeyryan42 (approved):

josephgregoryii (approved):
WOO! :shipit:

PR #299, opened 2022-07-28
More image improvements

aswanson-nr:
# Summary
I noticed that we were still getting dinged for the images we load "below the fold". As I dug into that I noticed that all of the SVGs in the catalogue were being loaded "eagerly" and that the landing pages were not using the version of the image processed by sharp during build.

This PR does the following:
* Implements a generic `SafeImage` component that uses `GatsbyImage` and falls back to `img` if needed
* Lazy load all SVGs
* Update `QuickstartImg` to use the new image component
* Update dashboard screenshots carousel to use the new image component
* Update landing banner image to use the new image component
* Various linting fixes

### DIFF
--- a/src/components/LandingBanner/index.js
+++ b/src/components/LandingBanner/index.js
@@ -9,11 +9,12 @@ import { quickstart } from '../../types';
 import QuickstartImg from '../QuickstartImg';
 import defaultImage from '../../images/defaultQuickstartImage.png';
 import BannerBackground from './BannerBackground';
+import SafeImage from '../SafeImage';
 
 const IMAGE_DISPLAY_BREAKPOINT = '1248px';
 
 const LandingBanner = ({ quickstart, className, location }) => {
-  const bannerImg = useDetermineBannerImg(quickstart, defaultImage);
+  const bannerImg = determineBannerImg(quickstart, defaultImage);
 
   const breadcrumbs = [
     {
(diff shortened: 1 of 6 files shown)


aswanson-nr:
This should wait on the fixes coming in for previews to make sure it doesn't break anything

src/components/LandingBanner/index.js
```diff
@@ -246,6 +260,19 @@ export const fragmentQuery = graphql`
...
+        childImageSharp {
+          gatsbyImageData(
+            placeholder: BLURRED
+            layout: CONSTRAINED
+            formats: [AUTO, WEBP]
+            height: 272
+          )
+        }
+      }
+    }
```
zstix, line 275: After some discussion, I think we probably _don't_ want to include this fragment. Since we're saying that we expect a `quickstart`.
```suggestion
```

src/components/LandingBanner/index.js
```diff
@@ -208,7 +220,9 @@ function useDetermineBannerImg(quickstart, defaultImage) {
     let image = bannerImg;
 
     for (const screenshot of quickstart.dashboards[0]?.screenshots ?? []) {
-      const { width, height } = await getURLMeta(screenshot);
+      const { width, height } =
```
zstix, line 223: We might want to add a default for these values.

Either way, and this is outside the scope of this PR, there is the possibility to throw a browser console error if the images fail to load for whatever reason (e.g. a `404`).

mickeyryan42 (approved):

josephgregoryii (approved):

josephgregoryii:
I see the lazy loading working!

PR #301, opened 2022-08-02
TS Config

josephgregoryii:


tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "lib": ["dom", "esnext"],
+    "jsx": "react",
+    "module": "esnext",
+    "moduleResolution": "node",
+
+    "resolveJsonModule": true,
```
josephgregoryii, line 9: allows JSON imports inside files

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+
+    "allowJs": true,
+
+    "declaration": true,
+    "declarationMap": true,
+    "declarationDir": "./src/types",
+    "emitDeclarationOnly": true,
```
josephgregoryii, line 16: [Emit d.ts files](https://www.typescriptlang.org/tsconfig#declaration)

Curious if we want this or not!

tsconfig.paths.json
```diff
@@ -0,0 +1,12 @@
+{
```
josephgregoryii, line 1: separated for import aliases

tsconfig.json
aswanson-nr: I'm not sure we need this since we don't publish any packages. Since it's done at build time, I don't think it provides any value for us.

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+
+    "esModuleInterop": true,
+
+    "strict": true,
+    "skipLibCheck": true
+  },
+  "include": ["src/"],
```
aswanson-nr, line 23: Will this allow us to make the gatsby files typescript? Since they live at the top level, not under `src/`

tsconfig.paths.json
aswanson-nr, line 1: It looks like we can add this configuration to the main `tsconfig`, looking at the [docs](https://www.typescriptlang.org/tsconfig#paths)

zstix (approved):

aswanson-nr (approved):

PR #303, opened 2022-08-03
NR-35139: Add ability to set custom meta titles and descriptions

mickeyryan42:
https://issues.newrelic.com/browse/NR-35139

The `description` prop doesn't seem to be used in `IOSeo.js`.  Can we remove it?

I am depending on the slug being the first part of the url which appears to work.  Please let me know if theres a case where this isn't true.

I did some reading on character limits for meta title/descriptions. It seems the limit for title is ~70 and description ~160. However these are apparently automatically truncated if they are too long, so I did not do any manual truncating of the strings and wanted to leave this open for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,14 +95,23 @@ To add a new set of meta tags for a quickstart, add the following to the quickst
 
 ```
 {
-  "slug": {
+  "<quickstart slug>": {
     "title": <title which should be in title meta tag>,
     "description": <description which should be in description meta tag>
   }
 }
 ```
 
-Create a pull request following the guidelines from [CONTRIBUTING.md](CONTRIBUTING.md)
+For example, to add a custom meta title and description to a quickstart with the slug cool-qs:
+
+```
+{
+  "cool-qs": {
+    "title": "Cool QS",
+    "description": "Cool qs description for SEO purposes"
+  }
+}
+```
 
 ### Environment variables
 
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ This can also be kicked off manually via a workflow run.
 ## Adding custom meta tags
 
 Custom meta tags can be added for a quickstart in [quickstart-metadata.json](src/data/quickstart-metadata.json)
-You will need to add a key to the object within this file in the following format:
+To add a new set of meta tags for a quickstart, add the following to the quickstart-metadata.json file.
 
 ```
 {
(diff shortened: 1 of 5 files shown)


src/data/quickstart-metadata.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "slug": "example",
+    "title": "meta title",
+    "description": "meta description"
+  }
+]
```
aswanson-nr, line 7: To make it easier to look up a quickstart, we could use the slug as the key.
```suggestion
{
  "example": {
    "title": "meta title",
    "description": "meta description"
  }
}
```

README.md
```diff
@@ -72,17 +83,33 @@ Runs every day to get various pages that relate to each quickstart from our vend
...
+
+```
+{
+  "slug": {
+    "title": <title which should be in title meta tag>,
+    "description": <description which should be in description meta tag>
+  }
+}
+```
+
```
aswanson-nr, line 104: I think we need a little more example-wise here. Maybe something like this:

Custom meta tags can be added for a quickstart in [quickstart-metadata.json](src/data/quickstart-metadata.json)
To add a new set of meta tags for a quickstart, add the following to the `quickstart-metadata.json` file.
```js
{
  "<quickstart slug>": {
    "title": <title which should be in title meta tag>,
    "description": <description which should be in description meta tag>
  }
}
```

For example, to add a custom meta title and description to a quickstart with the slug `cool-qs`:
```js
{
  "cool-qs": {
    "title": "Cool QS",
    "description": "Cool qs description for SEO purposes"
  }
}
```

src/data/quickstart-metadata.json
```diff
@@ -0,0 +1,6 @@
+{
+  "slug": {
+    "title": "meta title",
+    "description": "meta description"
+  }
```
aswanson-nr, line 5: As much as I'd like a placeholder here, I'd rather not end up with this matching a quickstart's slug eventually. I think the readme provides enough information to get started.
```suggestion
```

mickeyryan42: Yeah that makes sense, updated!

aswanson-nr (dismissed):

tabathadelane (approved):

aswanson-nr (approved):
Looks good!

PR #310, opened 2022-08-17
Re-reduce bundle size

aswanson-nr:
* We missed some imports from the default theme exporter file during PR
  review
* This lead to our bundle size being increase back to what it was before
  we worked on it
* This corrects those imports and adds an eslint rule to help us avoid
  this issue in the future
* Also renames QuickstartGrid to `js` instead of `jsx` so that eslint
  picks it up

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -66,5 +66,20 @@ module.exports = {
         ignoreRestSiblings: false,
       },
     ],
+    // This prevents us from accidently importing from the theme's `index` file
+    // and pulling in a bunch of extra dependecies.
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -66,5 +66,20 @@ module.exports = {
         ignoreRestSiblings: false,
       },
     ],
+    // This prevents us from accidently importing from the theme's `index` file
+    // and pulling in a bunch of extra dependecies.
```
aswanson-nr, line 70:
```suggestion
    // and pulling in a bunch of extra dependencies.
```

mickeyryan42 (approved):
I love this way to solve the issue!

tabathadelane (approved):

PR #313, opened 2022-08-18
Randomize featured quickstarts and instrument featured and popular

mickeyryan42:
https://issues.newrelic.com/browse/NR-39125

I was thinking it made sense to separate featured and popular into two HOC components vs. one reusable just to keep the logic decoupled, but this could easily be one wrapper as well.


src/components/withFeaturedInstrumentation.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
aswanson-nr, line 1: Let's combine these two higher order components into one and pass the tessen category in as a prop.
While I like the idea of separating these out into their own component, overall the only difference is the category and I think the duplication out weighs the explicitness here.

src/pages/index.jsx
```diff
@@ -229,13 +238,16 @@ const QuickstartsPage = ({ data, location }) => {
                 {!loadComplete && <Spinner />}
                 {loadComplete && (
                   <Slider {...indexSettings}>
```
aswanson-nr, line 240: It might be worth pulling the slider out into a component, that way we can encapsulate all this extra instrumentation code and reduce the footprint in the index page JSX

src/utils/shuffleArray.js
```diff
@@ -0,0 +1,15 @@
+const shuffleArray = (arr) => {
```
aswanson-nr, line 1: Let's get some unit tests on this

src/components/SliderWrapper.js
```diff
@@ -0,0 +1,43 @@
...
+import Slider from 'react-slick';
+import withInstrumentation from '@components/withInstrumentation';
+import QuickstartTile from '@components/QuickstartTile';
+// Utils
+import shuffleArray from '@utils/shuffleArray';
+// Types
+import { quickstart } from '../types';
+
+const SliderWrapper = ({ indexSettings, quickstarts, category }) => {
+  const shuffledQuickstarts = useMemo(() => shuffleArray(quickstarts), []);
```
josephgregoryii, line 15: Not a blocker, just wondering if we need to add a `quickstarts` dependency into the array?

josephgregoryii (dismissed):
Code looks solid to me - thanks for the walkthrough on friday!

src/components/SliderWrapper.js
mickeyryan42: Ah, so the issue is that if its tracking changes to quickstarts, when the component reloads `shuffleArray` runs again and shuffles the quickstarts causing the problem i mentioned where on hovering the tiles the order changes.  I'm using the memo so it memoizes the value and prevents it so I intentionally left it out because of that.  I wonder if I can do away with the empty array altogether

mickeyryan42: [screen-capture.webm](https://user-images.githubusercontent.com/105245033/185985805-16c654ee-6488-4759-818b-2e635b7bd6e1.webm)
Yeah it gives me the same to remove the array as well.  Heres a clip so yall can see what im referring to.

josephgregoryii: Very odd behavior! makes total sense :)

aswanson-nr: Let's get an eslint disable comment on this, possibly with a little explanation as to why we want to ignore the normal rules in this case!

src/components/SliderWrapper.js
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <Slider {...indexSettings}>
+      {category === 'MostPopularQuickstartClick' && <SuperTiles />}
```
aswanson-nr, line 19: If we ever need to change the logic around when these sliders show up, we'll have to go into this component and make changes as well as make changes to the index page. To reduce the footprint for future changes, could we pull this logic out to the index page?

aswanson-nr (commented):
Overall looks great! I just had 2 small comments

src/components/SliderWrapper.js
mickeyryan42: Can you check out the way i modified it? I just basically passed `showSuperTiles={true}` and use it in the component. I wasn't sure if there was another way you meant you'd like to handle it.  I was thinking we could conditionally pass it as a child from index and render props.children in SliderWrapper as an alternative

aswanson-nr: Hey @mickeyryan42, your changes look great! I was definitely a little confused as to what `<SuperTile />` is, now I see that it is the extra guided install tile.

src/components/SliderWrapper.js
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <Slider {...indexSettings}>
+      {showSuperTiles && <SuperTiles />}
+      {shuffledQuickstarts.map((pack, i) => {
+        const InstrumentedTile = withInstrumentation(QuickstartTile);
+
+        return (
+          <InstrumentedTile
+            category={category}
+            index={i}
```
aswanson-nr, line 37: Now that I finally get what `<SuperTiles />` is doing here, should we do `i+1` here when the super tile is added to the slider? Otherwise, the indexing will be off by one since the super tile takes the first spot in the slider.

mickeyryan42: Thats a great call.  I didn't actually know what SuperTiles was doing until i mistakenly added it to the featured section and couldnt figure out where the guided install tile was coming from lol. I'm going to update real quick with that change

src/components/withInstrumentation.js
```diff
@@ -0,0 +1,33 @@
...
+  InstrumentedComponent.propTypes = {
+    name: PropTypes.string.isRequired,
+    index: PropTypes.number.isRequired,
+    category: PropTypes.string.isRequired,
+  };
+
+  return InstrumentedComponent;
+};
+
+export default withFeaturedInstrumentation;
```
aswanson-nr, line 33:
```suggestion
export default withInstrumentation;
```

aswanson-nr (approved):

josephgregoryii (approved):
:shipit:

PR #314, opened 2022-08-23
Search ranking

aswanson-nr:
# Summary

- Adds search ranking
  - Ranking is defined by a list of rules that check for things like exact string match, word match, substring match, etc.
  - These rules are defined in priority order
- Make eslint ignore generated gatsby types

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "eslint-plugin-markdown": "^2.2.1",
     "jest": "^26.4.0",
     "prettier": "2.2.1",
+    "ts-jest": "^26.4.0",
     "tsc": "^2.0.4",
     "typescript": "^4.7.4"
   }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -53,6 +53,7 @@ module.exports = {
     '**/node_modules/**/*',
     '**/public/**/*',
     '**/static/*',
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -76,6 +76,7 @@
     "eslint-plugin-markdown": "^2.2.1",
     "jest": "^26.4.0",
     "prettier": "2.2.1",
+    "ts-jest": "^26.4.0",
```
josephgregoryii, line 79: I noticed the test file in this PR is still javascript - is this for possible future development for ts-jest?

josephgregoryii (approved):
Code looks cleaaaaaannnnn. :shipit:

mickeyryan42 (approved):
This looks awesome! The only thing i can think of is whether we want a comment on the regex in searchRanking.ts.  It's pretty self explanatory given the fn name so I don't think its a big deal either way.  Giving it an approval

package.json
aswanson-nr: This was needed to get jest to run tests that use `.ts` files, so it was required even if the jest files are `.js`.
We use it over in the [quickstarts repo](https://github.com/newrelic/newrelic-quickstarts/blob/main/utils/package.json#L47)

PR #317, opened 2022-08-26
NR-41180: Marketing changes

mickeyryan42:
https://issues.newrelic.com/browse/NR-41180

This PR:
1. Disables the click from the New Relic logo in the header when `utm_medium` is set.
2. Hides the header navigation when `utm_medium` is set
3. Adds the entire urlParams object to instrumentation for install button.

### DIFF
--- a/src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
+++ b/src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
@@ -83,8 +83,7 @@ const GlobalHeader = ({ className, activeSite }) => {
   const UserIsInMainPage = location.pathname === '/instant-observability/';
   const showGetStarted = !!UserIsInMainPage;
   const [isOpen, setOpen] = useState(false);
-  const urlParams = new URLSearchParams(location.search);
-  const utmCode = urlParams.get('utm_medium');
+  const [utmCode, setUtmCode] = useState(true);
 
   useEffect(() => {
     if (isOpen) {
@@ -93,6 +92,13 @@ const GlobalHeader = ({ className, activeSite }) => {
       document.body.style.overflowY = 'visible';
     }
   }, [isOpen]);
+
+  useEffect(() => {
+    const urlParams = new URLSearchParams(location.search);
+    const utmCodeParam = urlParams.get('utm_medium');
+
+    setUtmCode(utmCodeParam);
+  }, [location.search]);
 
   useEffect(() => {
     if (isOpen && hasChangedPage) {
@@ -158,10 +164,11 @@ const GlobalHeader = ({ className, activeSite }) => {
               align-items: center;
               outline: none;
               ${utmCode &&
+              !UserIsInMainPage &&
               `
-                pointer-events: none;
-                cursor: default;
-              `}
+                    pointer-events: none;
+                    cursor: default;
+                  `}
               @media screen and (max-width: ${NAV_BREAKPOINT}) {
                 width: 7.5rem;
               }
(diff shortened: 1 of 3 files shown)


src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
```diff
@@ -83,6 +83,8 @@ const GlobalHeader = ({ className, activeSite }) => {
   const UserIsInMainPage = location.pathname === '/instant-observability/';
   const showGetStarted = !!UserIsInMainPage;
   const [isOpen, setOpen] = useState(false);
+  const urlParams = new URLSearchParams(location.search);
+  const utmCode = urlParams.get('utm_medium');
```
aswanson-nr, line 87: What happens if a user hits the main (catalogue) page with a utm code? I think we may want to ensure that we aren't on the catalogue page as part of this logic.

aswanson-nr (commented):
Looks good to go! I noticed one small edge case, but it might be fine as is!

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
mickeyryan42: Thats a good call, I added some logic to only do so on the landing pages.  

I also noticed for some reason the conditional css wasn't working, which I thought it had been previously.  I went ahead and added the utmCode to state and verified its now working.  The reason im passing `true` as an initial value to state is because otherwise we see a flash of the nav before it gets hidden.  This way its hidden initially and only shows up after state is set so theres no point where the nav shows then disappears.

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
```diff
@@ -251,7 +264,10 @@ const GlobalHeader = ({ className, activeSite }) => {
                 }
               `}
             >
-              {createNavList('main', activeSite)}
+              {UserIsInMainPage ||
+                (!UserIsInMainPage &&
+                  !utmCode &&
+                  createNavList('main', activeSite))}
```
aswanson-nr, line 270: I'm still not seeing the nav links on the catalogue page when the utm_medium code is present. I think it's something to do with the boolean logic here. Since the logic here is getting a little more complex, what about moving it over to a function that decides whether or not to display the nav links?

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
aswanson-nr: Good call on that! It definitely looks better when the don't appear then disappear. However, this does change the behavior of the normal landing page, because now the links will "pop-in" on those pages. Personally, I think this is fine, but we should probably run it by the team beforehand!

src/@newrelic/gatsby-theme-newrelic/components/GlobalHeader.js
mickeyryan42: i just noticed the paren is after createNavList instead of utmCode so its improperly grouped.  Updating that

mickeyryan42:
@aswanson-nr I had some issues with prettier formatting the parens in ways that broke the logic but i think it should be ready to go

zstix (approved):

aswanson-nr (approved):
This looks good to go!!
I have one quality of life comment, but it's definitely not required to finish this up!

There are 3 locations in this JSX where we are using boolean logic to determine what to display or hide, but the logic itself does not indicate why it is showing/hiding the content.
To clear this up a bit for our future selves, we could make a variable to house the boolean logic.
```js
const isMarketingLandingPage = utmCode && !UserIsInMainPage;
```

PR #318, opened 2022-08-26
Fix guided install showing up twice

aswanson-nr:
# Summary

Stops the guided install tile from showing up more than once on the page. Also fixes a few linting issues.


tabathadelane (approved):
I love it!

zstix (approved):

PR #320, opened 2022-08-31
Guided install design changes

mickeyryan42:
https://issues.newrelic.com/browse/NR-43198

Changing the text to 18px to match the other tiles works without  a word break/ellipsis until about 311px screen width.  If we want we can maybe ellipsis at a breakpoint here but the smallest screen size we should encounter is 320px.

Since `GuidedInstallTileMostPopular` was the one being used, I deleted the other and renamed it `GuidedInstallTile`.  This caused a weird diff but the main changes are:

- Added hover css to the tile
- Added the onClick to the tile itself
- Removed the css to break word and ellipsis the text

**Note**
The `GuidedInstallTile` component has an event called clickSuperTile.  Since we are removing SuperTiles, should we rename this or would that cause problems with tracking events?


tabathadelane:
Added the hover feels so much better. These are nitpicks, but I'm curious what the team thinks:
Screen Shot 2022-08-31 at 9 17 36 AM
Is there enough difference between the main text and the "install New Relic" text with this change? 


Screen Shot 2022-08-31 at 9 17 54 AM
Did we know there is a spinner on this tile when you click it?

src/components/GuidedInstallTile.js
```diff
@@ -30,58 +38,156 @@ const GuidedInstallTile = () => {
...
+            border-bottom: 1px solid var(--primary-text-color);
+            &:hover {
+              background: none;
+              color: var(--system-text-muted-light);
+            }
+          `}
+        >
+          Install New Relic
+        </Button>
+        {startNavigation && (
```
aswanson-nr: Why do we need a spinner here? This seems like something we can remove

aswanson-nr (commented):
Looks good to me! But can we get rid of that spinner?

PR #326, opened 2022-09-01
Guided install design changes

mickeyryan42:
Closed and reopened because of gatsby build cache failure

https://issues.newrelic.com/browse/NR-43198


src/components/GuidedInstallTile.js
```diff
@@ -1,87 +1,164 @@
...
-    navigate(platformUrl);
-  };
+        &:hover {
+          color: initial;
+        }
+      `}
+      href={
+        isReturningUser
+          ? getGuidedInstallStackedNr1Url(NR1_GUIDED_INSTALL_NERDLET)
+          : SIGNUP_LINK
```
aswanson-nr, line 30: I think this is will always evaluate to the `SIGNUP_LINK` since `isReturningUser` will always be false during server side rendering. The logic worked before because it was run when the user clicks the button.

aswanson-nr (commented):
Looks good! Just noticed an issue where it will always send users to the signup page.

src/components/GuidedInstallTile.js
aswanson-nr:
```js
const [link, setLink] = useState(SIGNUP_LINK);

useEffect(()=> {
  const isReturningUser = Boolean(Cookies.get('ajs_user_id'));
  const loc = isReturningUser ? getGuidedInstallStackedNr1Url(NR1_GUIDED_INSTALL_NERDLET) : SIGNUP_LINK;
  setLink(loc)
}, [])
```

mickeyryan42: Updated to just use the SIGNUP_LINK

tabathadelane (approved):
Looks good to me. 👍

aswanson-nr (approved):

tabathadelane (approved):

PR #331, opened 2022-09-12
Add partner field to tessen events

mickeyryan42:
https://issues.newrelic.com/browse/NR-47908

### DIFF
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -175,7 +175,7 @@ const InstallButton = ({
       quickstartUrl: quickstart.packUrl,
       super_tiles_treatment: treatment,
       urlParameters: Object.entries([...parameters]),
-      partner: true,
+      partner: quickstart.keywords.includes('newrelic partner'),
       quickstartButtonText: hasInstallableComponent
         ? 'Install quickstart'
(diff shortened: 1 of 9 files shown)


aswanson-nr (commented):
Hey @mickeyryan42, the AC for this ticket was not correct. We actually only want to add the event if it's a partner quickstart. I added some more details to the ticket!

josephgregoryii (approved):
Looks good to me!

zstix (dismissed):

src/components/InstallButton.js
```diff
@@ -175,6 +175,7 @@ const InstallButton = ({
       quickstartUrl: quickstart.packUrl,
       super_tiles_treatment: treatment,
       urlParameters: Object.entries([...parameters]),
+      partner: quickstart.keywords.includes('newrelic partner'),
```
aswanson-nr, line 178: It could be worth defining a `isPartner` function that takes in a quickstart, that way if the logic ever changes, we only need to update it in one place.

aswanson-nr (approved):

josephgregoryii (approved):

aswanson-nr (approved):

PR #342, opened 2022-11-04
fix: update theme version to 6.13.6

clarkmcadoo:
This incorporates a Tessen fix along with a Gatsby gql fix


tabathadelane (approved):
great work. looks awesome.

PR #346, opened 2022-12-02
NR-67849 Always show install now CTA

mickeyryan42:
https://issues.newrelic.com/browse/NR-67849


mickeyryan42:
Since quickstarts should always have data sources, closing and ticket blocked until discussion with Becky

PR #347, opened 2022-12-20
Clark/add nerdgraph call

clarkmcadoo:
There were several functions that needed to be updated to no longer be
async, we were creating a condition where a Promise was being passed vs
a boolean.  Also, we removed a useEffect that was setting the params via
state and just used a const instead.

### DIFF
--- a/src/components/InstallButton.js
+++ b/src/components/InstallButton.js
@@ -41,12 +41,32 @@ const checkUtmParameters = (parameters) => {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


clarkmcadoo:
to QA: 

1. log into one.newrelic.com
2. go to an Install page and click the install button (i.e. java as it has an installable component), It should go to the appropriate install page in the product
3. Log out of one.newrelic.com
4. repeat step 2, it should go to newrelic.com/signup

LizBaker (dismissed):

LizBaker (approved):

PR #350, opened 2023-01-11
Remove Codestream shift to top

tabathadelane:
This was causing a bug in Firefox that caused the urls that showed on hover of a quickstart to off by one - they still navigate to their correct pages though. 

However after some investigation, we noticed that only Firefox had Codestream being forced to the front of the Quickstarts at all and we learned that the sorting function was the root of this bug. Coincidentally, we think marketing no longer wants Codestream in the front so this has gone unnoticed until now. 

This PR also removes the code that shift codestream to the top.


sunnyzanchi (approved):
good call breaking this into two commits in the same PR!

LizBaker (approved):

PR #376, opened 2023-05-08
Support data sources as install options

aswanson-nr:
# Summary

Adds data sources into the query that pulls quickstarts from Nerdgraph. Also updates the install button to head into the production when a quickstart has data sources, but no install plans.


josephgregoryii (approved):

aswanson-nr:
We will be building support for this into a future version. Closing this PR

